PR #411, opened 2014-01-09
Backup os user

zipkid:
Allow configuration of user and group performing the actual backup.


mhaskel:
@zipkid thanks for the contribution, but we haven't heard from you in a while so we're going to have to close this PR for now. If this is still something you're interested in adding to the module, please address the comments and submit a new PR. Thanks again!

PR #427, opened 2014-01-24
Allow override_options set to undef to completely remove the corresponding key reverting to the mysql default

radford:
This is on top of #426.

### DIFF
--- a/templates/my.cnf.erb
+++ b/templates/my.cnf.erb
@@ -8,7 +8,7 @@
 <%         vi.each do |vii| -%>
 <%=          ki %> = <%= vii %>
 <%         end -%>
-<%       else -%>
+<%       elsif vi != :undef -%>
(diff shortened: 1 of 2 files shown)


templates/my.cnf.erb
```diff
@@ -1,21 +1,18 @@
...
+[<%=   k %>]
+<%     v.sort.map do |ki, vi| -%>
+<%       if vi == true or v == '' -%>
+<%=        ki %>
+<%       elsif vi.is_a?(Array) -%>
+<%         vi.each do |vii| -%>
+<%=          ki %> = <%= vii %>
+<%         end -%>
+<%       elsif vi != :undef -%>
+<%=        ki %> = <%= vi %>
```
igalic: do i understand this code correct that  `foo = ''` will be set, but `foo = undef` will be unset?

PR #548, opened 2014-07-30
Add support for MariaDB > 10.0.0

igalic:
MariaDB 10.0.0 allows for usernames up to 80 bytes, removing a long-standing restriction in MySQL.

We introduce two functions to mysql.rb: `version_check` and `username_length`.
version_check wraps `Puppet::Util::Package.versioncmp(product_version, version)`, and allows us to compare the version of mysql we're running under, with a few mile-stones, which allow us to decide its feature-set. `username_length` in turn will return the length based on those versions.

Why?
----

mysql will happily accept a username that doesn't fit into its column, truncating it. This works "fine" for a system that is one-shot, but in a system that's supposed to be idempotent, we'll **actually** notice that something is amiss.

This PR addresses [MODULES-151](https://tickets.puppetlabs.com/browse/MODULES-151)

### DIFF
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -1,9 +1,27 @@
+require 'puppet/util/package'
+
 class Puppet::Provider::Mysql < Puppet::Provider
 
   # Without initvars commands won't work.
   initvars
   commands :mysql      => 'mysql'
   commands :mysqladmin => 'mysqladmin'
+
+  # if we actually *can* find mysql in the path, we check the version, and
+  # based on that, we check if the passed version is >= the product_version
+  def self.version_check(version)
+    begin
+      # mysql --version # for mysql and mariadb
+      # =#> mysql  Ver 14.14 Distrib 5.5.37, for debian-linux-gnu (x86_64) using readline 6.2
+      # =#> mysql  Ver 15.1 Distrib 10.0.12-MariaDB, for debian-linux-gnu (x86_64) using readline 5.1
+
+      product_version = mysql('--version').split[4].chop
+    rescue Puppet::ExecutionFailure
+      false
+    else
+      true if Puppet::Util::Package.versioncmp(product_version, version) >= 0
+    end
+  end
 
   # Optional defaults file
   def self.defaults_file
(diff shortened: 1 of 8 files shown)


mhaskel:
Hey @igalic, just wondering if this is still on your radar...

igalic:
holy wow yes no…
i still don't know how to

wondering if @DavidS has any ideas :P

DavidS:
Types cannot represent provider specific restrictions. e.g. when running on the master, the type has no idea which provider to use (by default).

I liked the idea with a feature, but reading the source code, it seems to me that the features on types are only used for documentation purposes. Which only further supports that there is no way to bubble up information from the provider.

Given that basic restriction, I believe the best we can do is keep the long_usernames feature for documentation purposes and throw the error on the client when the provider actually sees the value the first time. This could be enhanced by an *optional* call into the provider if it is available, e.g. because the user has set it explicitly. The mysql module could even support this by setting the providers explicitly in mysql::server::providers if it wants to be extra cool. I'm not convinced it is worth the effort though ...


To get "proper" compile-time validation, one would need to do something like surfacing the mysql version in a fact and use that to influence the validation in the type. This doesn't work for two reasons: a) in the most important use-case of installing-and-configuring mysql on a new node the fact is empty and b) it would prevent configuring both mysql and mariadb on the same host.

igalic:
i don't think b) is a valid use-case for this module

igalic:
@mhaskel  ping re @DavidS' a)

DavidS:
Thinking more about it, I realise that the user already has to select *which* DB/version to install, no? This information is - by definition - available during the catalog compile. Can we get at that reliably?

Without looking at the code, in the worst case, we could create separate mysql/mariadb types which share implementation, wrapped in a multiplexor define. *shudder*

igalic:
the problem with that information being available is that it's available through the classes. so in the types/providers we'd either have to duplicate the logic (lolnope), or else try and access the information made available in the class from within the type/provider.

the huge disconnect between classes (and defined types), and the types&providers system is sometimes infuriating >_<
****
btw, i've *tried* what you suggest. a long long time ago. but, it kinda felt unsatisfactory.
but, a long long time ago, i didn't know enough about types/providers, so maybe another attempt would yield better results.

mhaskel:
o_o

This is all very complicated and making my head hurt.

So, basically, I'm not sure what the best bet is here :|

igalic:
closing in favour of #722. Which will *not* cover the "corner-case" of working on the first run… because, facter…

PR #549, opened 2014-08-01
Added support for backing up triggers and routines to the mysql::server:backup module

stevesaliman:
The mysqldump command issued by the module-generated mysqlbackup.sh script does not include triggers or routines when it does a file-per-database backup.  This means that if users try to restore a database using one of theses backup files, they will not be restoring the full structure of their database, just the data, which can leave users with broken applications due to missing routines.   I think that it is important to support triggers and routines in a backup, especially if users are only backing up selected database, in which case the triggers and routines will not get backed up anywhere else. 

To resolve this issue, I’ve added support for 2 new ```mysql::server::backup``` parameters (```mysql::backup``` is marked as deprecated, so I didn’t modify it) .  The first is called ```include_triggers``` and the second is called ```include_routines```.  When these parameters are true, the appropriate options are added to the mysqldump command.  These flags have no effect when we are doing a full backup to a single file, since the triggers and routines will be backed up with the information_schema.

The default value for both new options is true, reflecting my belief that omitting triggers and routines is an oversight and not a feature.  If you agree, this pull request is ready to go.

If, on the other hand, you feel that including triggers and routines should be an opt-in choice instead of default behavior, you’ll need to change the defaults in the module and fix some unit tests accordingly.

Either way, I think the ability to support the backup of triggers and routines in a file-per-database mode is important because it is the only way for users to generate a backup file for a single database that can be used to fully restore a single named database to its original state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,14 @@ Array of databases to specifically back up.
 
 Whether a separate file be used per database.
 
+#####`include_triggers`
+
+Whether or not to include triggers for a each database when doing a `file_per_database` backup. Defaults to `true`.
+
+#####`include_routines`
+
+Whether or not to include routines for each database when doing a `file_per_database` backup. Defaults to `true`.
+
 #####`ensure`
 
 Allows you to remove the backup scripts. Can be 'present' or 'absent'.
--- a/manifests/server/backup.pp
+++ b/manifests/server/backup.pp
@@ -12,6 +12,8 @@ class mysql::server::backup (
   $delete_before_dump = false,
   $backupdatabases = [],
   $file_per_database = false,
+  $include_triggers = true,
+  $include_routines = true,
   $ensure = 'present',
   $time = ['23', '5'],
   $postscript = false,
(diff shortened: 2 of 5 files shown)


mhaskel:
Hi @stevesaliman, thanks for the contribution, and apologies this fell off the radar. There were some major refactors of backups since this was submitted, could you rebase this and squash it down to a single commit? Thanks.

stevesaliman:
With the heavy refactoring, it might be easier to redo my changes from the latest master.  I see that the backups now support MySql Enterprise Backup and a Percona backup.  I didn't look into the Percona backups, but it doesn't look like mysqlbackup has any options for including or excluding triggers and routines.  I think this leads to one of 3 possible solutions:

1. I could add ```include_triggers``` and ```include_routines``` as configurable options like I did in the original pull request, but that means I'd need to also include them in the percona and mysqlbackup manifests, where they would be ignored.

2. I could add the new options to the mysqldump.pp manifest only, and require the user to do something with hieradata or the like to override the defaults.

3. I could simply hard code the --routines and --triggers options in the script that is calling mysqldump.  so that backups will always capture everything, which is what I suspect the other two backup technologies do.

What are your thoughts?

igalic:
i think we have no choice but to go for 1.

stevesaliman:
I've gone with option 1.  I also found a privilege bug that was preventing triggers from being backed up, which I've also fixed.  I'll create a new pull request from my feature branch, which was rebased to the master branch this morning.  All tests work, so I think we're good to go.

PR #562, opened 2014-09-04
Add new parameters create_root_user and create_root_my_cnf.

franzs:
This allows the galera module and others to write ${::root_home}/.my.cnf
independently from create the mysql user. This is useful for cluster
setups where you want to create ${::root_home}/.my.cnf on every node
but create the user only once.


mhaskel:
Closing this since #578 was merged

PR #567, opened 2014-09-14
Add OpenBSD support.

buzzdeee:
This is a first step in OpenBSD support, tested with mysql-server-5.1.73p0v0 package on OpenBSD 5.6-current, using the following configuration stored in hiera.
```yaml
mysql::server::root_password: topsecret
mysql::server::manage_config_file: true
mysql::server::purge_conf_dir: true
mysql::server::package_ensure: present
mysql::server::service_enabled: true
mysql::server::remove_default_accounts: true
mysql::server::users:
  'dashboard@localhost':
    ensure: present
    max_connections_per_hour: 0
    max_queries_per_hour: 0
    max_updates_per_hour: 0
    max_user_connections: 0
    password_hash: '*TOPSECRETTOPSECRETTOPSECRET'
mysql::server::grants:
  'dashboard@localhost/dashboard_production.*':
    ensure: present
    options:
      - GRANT
    privileges:
      - ALL
    table: 'dashboard_production.*'
    user: dashboard@localhost
mysql::server::databases:
  dashboard_production:
    ensure: present
    charset: utf8
mysql::server::override_options:
  mysqld:
    user: _mysql
    datadir: /var/mysql
```

The odds are that I have to use the mysql::server::override_options :(
Fixing that, is for another diff, in order to keep this one clean and easier to review.

The change to server/service.pp is because there is no 'mysql' user and group, but _mysql user and _mysql group.
Another solution would be to add a $mysqlgroup variable to mysql::params list and use that. 
However, the $mysql::params::log_error file is owned by _mysql:_mysql on my OpenBSD system instead of the root_group.

This is only a first step, that allowed me to create a database. 
 - Further steps I'd look into would be:
   * get rid of the currently required usage of override_options
   * introduce a $mysqluser and $mysqlgroup to params.pp and use those throughout the classes
   * in other places, i.e. the backup config, there the group root is hardcoded, but on OpenBSD there is no group root, but wheel, so use the $root_group there

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -10,10 +10,16 @@ class mysql::server::service {
     }
   }
 
+  if $mysql::server::override_options['mysqld'] and $mysql::server::override_options['mysqld']['user'] {
+    $mysqluser = $mysql::server::override_options['mysqld']['user']
+  } else {
+    $mysqluser = $options['mysqld']['user']
+  }
+
   file { $options['mysqld']['log-error']:
     ensure => present,
-    owner  => 'mysql',
-    group  => 'mysql',
+    owner  => $mysqluser,
+    group  => $::mysql::server::root_group,
   }
 
   service { 'mysqld':
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -251,6 +251,33 @@ class mysql::params {
       $daemon_dev_package_name     = undef
     }
 
+    'OpenBSD': {
+      $client_package_name = 'mariadb-client'
+      $server_package_name = 'mariadb-server'
+      $basedir             = '/usr/local'
+      $config_file         = '/etc/my.cnf'
+      $includedir          = undef
+      $datadir             = '/var/mysql'
+      $log_error           = "/var/mysql/${::hostname}.err"
+      $pidfile             = '/var/mysql/mysql.pid'
+      $root_group          = 'wheel'
+      $server_service_name = 'mysqld'
+      $socket              = '/var/run/mysql/mysql.sock'
+      $ssl_ca              = undef
+      $ssl_cert            = undef
+      $ssl_key             = undef
+      $tmpdir              = '/tmp'
+      # mysql::bindings
+      $java_package_name   = undef
+      $perl_package_name   = 'p5-DBD-mysql'
+      $php_package_name    = 'php-mysql'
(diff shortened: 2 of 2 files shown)


tphoney:
ping @buzzdeee this needs a rebase and the commits need squashed. Thanks for the effort you have put in.

buzzdeee:
Thanks for pinging, I'll take care as soon as I find time. ETOOMANYPULLREQUESTS ;)

tphoney:
I hear ya !!

buzzdeee:
squashed and rebased, Travis seems to be happy. 
Got a little merge conflict when rebasing in manifests/server/service.pp, but a test just showed, still seems to work for me. 

cheers,
Sebastian

igalic:
thanks a lot @buzzdeee!

manifests/server/service.pp
```diff
@@ -10,10 +10,16 @@
...
+  } else {
+    $mysqluser = $options['mysqld']['user']
+  }
+
   file { $options['mysqld']['log-error']:
     ensure => present,
-    owner  => 'mysql',
-    group  => 'mysql',
+    owner  => $mysqluser,
+    group  => $::mysql::server::root_group,
```
mhaskel: @buzzdeee @igalic The group change breaks idempotency on non-debian systems [(example)]( https://www.google.com/url?q=https%3A%2F%2Fjenkins-modules.puppetlabs.com%2Fview%2F1.%2520failing%2520jobs%2Fjob%2Fmodules_puppetlabs-mysql_intn-sys_ext-master-enterprise-stable%2F36%2FPLATFORM_CONFIG%3Dcentos-5-64mda%2CSLAVE_LABEL%3Dbeaker%2Fconsole&sa=D&sntz=1&usg=AFQjCNGvo20G6wjvdWodNjjMEqFtS3Asqg). I'm assuming this change was needed for OpenBSD support, but this should be refactored to leave the group `'mysql'` for other OSes.

buzzdeee: sorry for that, I'm looking into it

mhaskel: Thanks @buzzdeee .

buzzdeee: no problem, I introduced it, I should at least come up with a proposal to fix it. I kind of feel responsible for the sh*t I produce, ;) see #691

PR #586, opened 2014-10-24
(MODULES-1685) permit use of mysql::db without mysql::server

larsks:
This removes the hard dependencies on mysql::server from manifests/db.pp.  This permits one to use this module to manage databases in an existing server, possibly located on a remote host (with a properly configured /root/.my.cnf).

The dependencies are instead placed in mysql/server.pp and are implemented using collectors, so that the dependencies are only created if `mysql::server` is realized.

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -23,7 +23,7 @@ define mysql::db (
     charset  => $charset,
     collate  => $collate,
     provider => 'mysql',
-    require  => [ Class['mysql::server'], Class['mysql::client'] ],
+    require  => [ Class['mysql::client'] ],
   }
   ensure_resource('mysql_database', $dbname, $db_resource)
 
@@ -31,7 +31,6 @@ define mysql::db (
     ensure        => $ensure,
     password_hash => mysql_password($password),
     provider      => 'mysql',
-    require       => Class['mysql::server'],
   }
   ensure_resource('mysql_user', "${user}@${host}", $user_resource)
 
@@ -41,7 +40,10 @@ define mysql::db (
       provider   => 'mysql',
       user       => "${user}@${host}",
       table      => $table,
-      require    => [Mysql_database[$dbname], Mysql_user["${user}@${host}"], Class['mysql::server'] ],
+      require    => [
+        Mysql_database[$dbname],
+        Mysql_user["${user}@${host}"],
+      ],
     }
 
     $refresh = ! $enforce_sql
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -78,5 +78,8 @@ class mysql::server (
     Anchor['mysql::server::end']
   }
 
-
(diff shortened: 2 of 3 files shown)


manifests/server.pp
```diff
@@ -78,5 +78,8 @@
     Anchor['mysql::server::end']
   }
 
-
+Class['mysql::server'] -> Mysql_database <||>
+Class['mysql::server'] -> Mysql_user <||>
```
hunner: Instead of collectors, could you add autorequires for `Class['mysql::server']` to the `mysql_database` and `mysql_user` types? Adding collectors without filters could do something bad for anyone using virtual resources, like creating all databases on all mysql servers.

hunner:
Closes MODULES-1685

manifests/db.pp
```diff
@@ -23,7 +23,7 @@
     charset  => $charset,
     collate  => $collate,
     provider => 'mysql',
-    require  => [ Class['mysql::server'], Class['mysql::client'] ],
+    require  => [ Class['mysql::client'] ],
```
stamak: I would like to see here 
<pre><code>
require  => defined(Class['mysql::server'])? {
                   true  => [ Class['mysql::server'], Class['mysql::client'] ],
                   false => [ Class['mysql::client'] ]
                },</code></pre>

manifests/db.pp
```diff
@@ -31,7 +31,6 @@
     ensure        => $ensure,
     password_hash => mysql_password($password),
     provider      => 'mysql',
-    require       => Class['mysql::server'],
   }
```
stamak: I would add here 
<code>Mysql_database[$dbname] -> Mysql_user["${user}@${host}"]</code>
or 
<code>require => Mysql_database[$dbname]</code>
this define is called only with all parameters anyway

manifests/server.pp
stamak: It is not possible here due to the aim of patch: we want to use this define without defined class mysql::server(for remote server where mysql server is already installed)
I would offer to clean it up or at least leave as is

phaedrusquality:
I would like to be able to use this soon in a release.

nelg:
We've been playing with the concept of using this module to manage RDS instances.

The problem with the above, which we ran into, was that it still only supports exactly 1 database server.  When the databases are remote, it is quite possible that the could be more than one that we want to manage from a single instance running puppet.

So, ideally if support could be added to choose which .my.cnf file the module uses, for each resource, that would be ideal.

danieldreier:
I just ran into a situation trying to use puppet to manage RDS databases where I would have needed this. I'll use this as soon as it's merged.

igalic:
if somebody gives their :+1: for collector i'll rebase this patch, and add documentation that we're using unfiltered collection.

alternatively, i could also add a filter parameter

igalic:
closing in favour of #736

PR #589, opened 2014-10-31
fix dealing with empty passwords

jbrunner:
Empty passwords ("") are passed to mysql_password("") which results in
hash: *BE1BDEC0AA74B4DCB079943E70528096CCA985F8

Expected value for password is NULL

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -34,7 +34,7 @@ define mysql::db (
   }
   $user_resource = {
     ensure        => $ensure,
-    password_hash => mysql_password($password_hash),
+    password_hash => $password_hash,
     provider      => 'mysql',
     require       => Class['mysql::server'],
   }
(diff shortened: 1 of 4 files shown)


dveeden:
This is similar to #646. Only one of these should be merged.

manifests/db.pp
```diff
@@ -28,9 +28,13 @@
   }
   ensure_resource('mysql_database', $dbname, $db_resource)
 
+  case empty($password) {
+    true  : { $password_hash = '' }
+    false : { $password_hash = mysql_password($password) }
```
igalic: i think puppet lint (or the future parser?) might complain here about a missing default case.

igalic:
general: i'm not sure which of those two approaches is better.

dveeden:
I'm in favor of #646 as that mimics the behaviour of MySQL's PASSWORD().

mhaskel:
Closing since #646 was merged. Thanks for the contribution, though @jbrunner !

PR #590, opened 2014-10-31
make user and table parameter of mysql_grant() optional

jbrunner:
Because name must match user and table parameters,
extract user and table defaults from name.

Struggled on this while trying to grant several users once using ensure_resource.

This will now work:

    $user_grant = {
      privileges => [ ... ],
      provider   => 'mysql',
    }
    $users = [ 'me@localhost/table.*', 'me@10.10.%/table.*' ]
    ensure_resource('mysql_grant', $users, $user_grant)

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,8 +246,6 @@ grants => {
     ensure     => 'present',
     options    => ['GRANT'],
     privileges => ['SELECT', 'INSERT', 'UPDATE', 'DELETE'],
-    table      => 'somedb.*',
-    user       => 'someuser@localhost',
   },
 }
 ```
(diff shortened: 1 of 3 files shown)


DavidS:
Hi @jbrunner , thanks for the work you have put in to this pull request. However due to inactivity and the fact that this cannot be merged as is anymor this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PS: please see these links for help with hoisting this code to the current master:
Rebasing: https://git-scm.com/book/en/v2/Git-Branching-Rebasing
Squashing: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

PR #596, opened 2014-11-06
Create log-bin directory if it doesn't exist

NoodlesNZ:
It seems the MySQL doesn't like it when you use a log-bin directory that doesn't exist. If you add this before the mysql package is installed then it will fail (if the mysql user doesn't exist). If you add the directory after you run the mysql::server class then mysql will fail to start the first time, it will only be resolved on a subsequent puppet run.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -18,6 +18,18 @@ class mysql::server::config {
       purge   => $mysql::server::purge_conf_dir,
     }
   }
+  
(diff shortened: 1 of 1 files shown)


NoodlesNZ:
Looking at this the dirname() function uses ruby's File.dirname(). If you don't specify a path (i.e. just mysql-bin) then dirname will return .

I haven't tested to see if this causes problems, but I'll do a check for . and stop puppet from trying to manage a directory for this.

NoodlesNZ:
Should be addressed in PR #654

PR #617, opened 2014-12-07
(MODULES-1143) Add package_manage parameters

juniorsysadmin:
This patch adds a package_manage parameter for both mysql::server and mysql::client


mhaskel:
Thanks @juniorsysadmin !

PR #621, opened 2014-12-09
Support defining `log_error` in mysql::server

tOnI0:
On RedHat 7 the default provider has been set to `mariadb`. However, when one uses MySQL instead, the error log file is still enforced to MariaDB's default (/var/log/mariadb/mariadb.log) which parent folder is obviously not created by MySQL packages, and Puppet fails.

This commit allows setting the path to the error log file to avoid the failure.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -2,6 +2,7 @@
 class mysql::server (
   $config_file             = $mysql::params::config_file,
   $includedir              = $mysql::params::includedir,
+  $log_error               = $mysql::params::log_error,
   $install_options         = undef,
   $manage_config_file      = $mysql::params::manage_config_file,
   $old_root_password       = $mysql::params::old_root_password,
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -9,7 +9,7 @@ class mysql::server::service {
     }
   }
 
-  file { $mysql::params::log_error:
+  file { $mysql::server::log_error:
     ensure => present,
     owner  => 'mysql',
     group  => 'mysql',


tOnI0:
Well I could think of a dirty solution:
 * add a `provider` parameter to init.pp (and appropriate subclasses)
 * remove inheritance on params.pp in init.pp
 * include params.pp from within init.pp, and clone all the resulting parameters (eg. `$_package_name = $mysql::params::package_name`)

This involves setting a default value of `undef` to all parameters in init.pp and assign the actual default values later in the class via clone parameters, or the user defined value if not equal to `undef`.

It's kind of messy but I use this pattern a lot in [my tomcat module](https://github.com/tOnI0/aco-tomcat).

Should I submit a pull request or is it a no go?

igalic:
from what i gather, you're hinting at a ::globals pattern, like we use in [postgresql](https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/globals.pp)

tOnI0:
Yes sir, I didn't know there was a name for it. Thanks for the reference

timmooney:
I like where @tOnI0 is going with this and we need a fix for this too for RHEL 6 with the MariaDB from the Red Hat Software Collections Library (SCL).

My problem with this approach is that it's different from the one that mysql::server::install uses to find the datadir and basedir settings.  It would be nice if we could come up with a consistent mechanism for overriding these configuration bits.

mhaskel:
@tOnI0 thanks for the contribution, we resolved this with https://github.com/puppetlabs/puppetlabs-mysql/pull/678. Hope that helps!

PR #634, opened 2014-12-31
MODULES-1520 add test to assert it does not break

cyberious:


mhaskel:
Depends on #635 to actually work

cyberious:
635 merged

PR #635, opened 2015-01-01
MODULES-1520 - update username validation

mhaskel:
Don't fail on validation where the user isn't quoted with special
characters. The providers quote these strings by default.


mhaskel:
Fixes puppetlabs/puppetlabs-mysql#633

PR #630, opened 2014-12-23
handling manage_config_file properly

KlavsKlavsen:
This should work when manage_config_file => false.. untested though.. I have not yet had time to read up on how to test using the spec setup.. links appreciated.

PR #636, opened 2015-01-05
more flexible service require

Micsi:
This makes the recently merged service-require [PR #615](https://github.com/puppetlabs/puppetlabs-mysql/pulls615) work better with externally managed config-files, even stored in unusual places as well as with user-specified packages. See discussion with original author at [PR #615](https://github.com/puppetlabs/puppetlabs-mysql/pulls615).
Works well with package-manage param [PR #617](https://github.com/puppetlabs/puppetlabs-mysql/pull/617)
This is sort of an alternative to [PR #630](https://github.com/puppetlabs/puppetlabs-mysql/pull/630)

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -20,7 +20,7 @@ class mysql::server::service {
     name     => $mysql::server::service_name,
     enable   => $mysql::server::real_service_enabled,
     provider => $mysql::server::service_provider,
-    require => [ File['mysql-config-file'], Package['mysql-server'] ]
+    require  => [ File['mysql-config-file'], Package['mysql-server'] ]
   }
 
 }
--- a/spec/acceptance/types/mysql_user_spec.rb
+++ b/spec/acceptance/types/mysql_user_spec.rb
@@ -73,4 +73,14 @@ describe 'mysql_user' do
       end
     end
   end
+  context 'using resource should throw no errors' do
+    describe 'find users' do
+      it {
+        on default, puppet('resource mysql_user'), {:catch_failures => true} do |r|
+          expect(r.stdout).to_not match(/Error:/)
+          expect(r.stdout).to_not match(/must be properly quoted, invalid character:/)
+        end
+      }
+    end
+  end
 end


manifests/server/service.pp
```diff
@@ -20,7 +20,9 @@
     name     => $mysql::server::service_name,
     enable   => $mysql::server::real_service_enabled,
     provider => $mysql::server::service_provider,
-    require => [ File['mysql-config-file'], Package['mysql-server'] ]
+    # if you manage the config file outside of the module or want mysql provided by another package, you may still use this service
+    # provided you set the parameters config_file resp. package_name on the mysql::server class
+    require  => [ File[$mysql::server::config_file], Package[$mysql::server::package_name]]
```
hunner: Could you conditionalize the `File[$mysql::server::config_file]` dependency on `$mysql::server::manage_config_file` (see #630 for example)

Micsi: Why should we? Pls see the discussion on this at #630? In the proposed form it works well with a externally manged config files. I think this is the better alternative because MySQL still requires a config file to start.
This form of the require ensures that the config exists, when the server starts but even allows not standard locations of the file.
Because of this, this PR is an alternative to #630.

Micsi:
@mhaskel I did the rebase and pushed the results. Who is responsible to remove the "needs-rebase" label?

mhaskel:
@Micsi I think something went wrong with your rebase ... there aren't any commits there from you anymore and the PR is showing no changes

hunner:
It looks like the rebase went wonky? There are now no lines changed. Try this:
```shell
git checkout service-require
git reset --hard 6c48b1c
git checkout master
git remote add puppetlabs https://github.com/puppetlabs/puppetlabs-mysql.git
git remote update
git reset --hard puppetlabs/master
git checkout service-require
git rebase master
```
<fix your merge conflict here, then `git add manifests/server/service.pp` to add the changes>
```shell
git rebase --continue
git push -f origin service-require
```

And that should get this PR up to date with only one commit, no merge conflicts, and the changes you want.

mhaskel:
Actually, it looks like this PR won't actually fix the same usecase as #630. If $manage_config_file is set to false this will be adding dependencies on resources that likely don't exist.

Micsi:
@mhaskel what added dependencies do you think of?
MySQL needs a config file in place to start. manage_config_file tells the module to mange it or not. The dependency to the config file has nothing to do with that, it is required anyway. With this PR the user is free to mange the config outside of the module and to store it in unusable places and still use the module service.

I for example want to copy an existing config file over to the server and use the module service. To require the config to exist before starting the machine is correct.

I think the management of the config file has nothing to do with the service. My version covers most use cases for configs.

mhaskel:
@Micsi We think something more along the lines of #630 is better, just not adding that dependency if the config file isn't being managed. If the config file is being managed externally, there should probably be a `before => Service['mysqld']` where the file is being managed, rather than the `require` in the `service`.

Micsi:
ok then. I'll do as you say ;-)

igalic:
@Micsi any updates here?

Micsi:
sorry, not yet. Will push requested changes asap.

igalic:
thank you everyone! this was fixed in #672

PR #637, opened 2015-01-10
Do the right thing when fqdn==localhost

dveeden:
Make sure not to declare root@localhost and @localhost if the fqdn is 'localhost'

```
Error: Duplicate declaration: Mysql_user[root@localhost] is already declared in file /etc/puppet/modules/mysql/manifests/server/root_password.pp:11; cannot redeclare at /etc/puppet/modules/mysql/manifests/server/account_security.pp:11 on node localhost
Error: Duplicate declaration: Mysql_user[root@localhost] is already declared in file /etc/puppet/modules/mysql/manifests/server/root_password.pp:11; cannot redeclare at /etc/puppet/modules/mysql/manifests/server/account_security.pp:11 on node localhost
```

Also remove root@localhost.localdomain and @localhost.localdomain in that case

### DIFF
--- a/manifests/server/account_security.pp
+++ b/manifests/server/account_security.pp
@@ -1,18 +1,34 @@
 class mysql::server::account_security {
   mysql_user {
-    [ "root@${::fqdn}",
-      'root@127.0.0.1',
+    [ 'root@127.0.0.1',
       'root@::1',
-      "@${::fqdn}",
       '@localhost',
       '@%']:
     ensure  => 'absent',
     require => Anchor['mysql::server::end'],
   }
-  if ($::fqdn != $::hostname) {
-    mysql_user { ["root@${::hostname}", "@${::hostname}"]:
+  if ($::fqdn != 'localhost.localdomain') {
+    mysql_user {
+      [ "root@localhost.localdomain",
+        "@localhost.localdomain"]:
       ensure  => 'absent',
       require => Anchor['mysql::server::end'],
+    }
+  }
+  if ($::fqdn != 'localhost') {
+    mysql_user {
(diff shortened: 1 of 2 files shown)


manifests/server/account_security.pp
```diff
@@ -1,14 +1,22 @@
...
       'root@::1',
-      "@${::fqdn}",
       '@localhost',
+      'root@localhost.localdomain',
+      '@localhost.localdomain',
       '@%']:
     ensure  => 'absent',
     require => Anchor['mysql::server::end'],
   }
+  if ($::fqdn != 'localhost') {
```
hunner: What if fqdn is localhost.localdomain? It would create duplicates in that case.

manifests/server/account_security.pp
```diff
@@ -1,14 +1,22 @@
 class mysql::server::account_security {
   mysql_user {
-    [ "root@${::fqdn}",
-      'root@127.0.0.1',
+    [ 'root@127.0.0.1',
       'root@::1',
-      "@${::fqdn}",
       '@localhost',
```
hunner: Do we need `root@localhost` too?

hunner:
Could you squash this, update as per comments, and add rspec-puppet tests for the cases of fqdn being localhost, localhost.localdomain, and neither? Thanks!

manifests/server/account_security.pp
dveeden: root@localhost should not be removed. That's the main account which is used if you connect over a UNIX socket, which is the default.

dveeden:
I've added some tests. Please check them as I'm not too familiar with rspec.

PR #640, opened 2015-01-11
Support for authentication plugins

dveeden:
Besides the default mysql_native_password there are more plugins available:
* mysql_old_password (pre 4.1 passwords, support for it will be dropped in MySQL 5.7)
* sha256_password (since 5.6?, more secure and stored a salted sha256 hash in authentication_string (instead of storing a non salted SHA1 hash in the password field)
* unix socket authentication
* PAM authentication
* Windows authentication
* No login (similar to /usr/sbin/nologin)

This PR makes it possible to define users which are authenticated by an authentication plugin.

Possible improvements:
* Support CREATE USER xxx IDENTIFIED WITH yyy USING zzz
* It seems like MariaDB 10.x+ uses IDENTIFIED VIA instead if IDENTIFIED WITH
* There is no good way to change the authentication plugin (See https://bugs.mysql.com/bug.php?id=67449 for details)
* For sha256_password old_password=2 must be set if you call PASSWORD()
* The authentication plugin is not loaded automatically

See also:
https://dev.mysql.com/doc/refman/5.7/en/authentication-plugins-available.html
https://mariadb.com/kb/en/mariadb/documentation/plugins/unix_socket-authentication-plugin/
https://mariadb.com/kb/en/mariadb/documentation/plugins/pam-authentication-plugin/

### DIFF
--- a/lib/puppet/provider/mysql_user/mysql.rb
+++ b/lib/puppet/provider/mysql_user/mysql.rb
@@ -12,13 +12,14 @@ Puppet::Type.type(:mysql_user).provide(:mysql, :parent => Puppet::Provider::Mysq
     # To reduce the number of calls to MySQL we collect all the properties in
     # one big swoop.
     users.collect do |name|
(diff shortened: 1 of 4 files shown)


hunner:
Thanks! Could you squash this into a single commit, or multiple if that would be more apt?

dveeden:
I've squashed everything into one commit

PR #641, opened 2015-01-19
Add type & provider for managing plugins

dveeden:
The basics for working with MySQL plugins works.

Some plugins like auth_socket and validate_password are shipped with the server. The enterprise edition also has a auditlog, encryption and threadpool plugin. There are also third-party plugins which add things like JSON parsing, etc.


spec/acceptance/types/mysql_plugin_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          ensure => present,
+          soname => 'auth_socket.so',
+        }
+      EOS
+
+      apply_manifest(pp, :catch_failures => true)
+    end
+
+    it 'should find the plugin' do
+      shell("select plugin_name from information_schema.plugins where plugin_name='auth_socket'") do |r|
```
cmurphy: I don't think this would work, since shell() shells out to the sh shell, not the mysql shell. Should look more like https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/acceptance/types/mysql_user_spec.rb#L27

cmurphy:
Could you add a section for this to the README?

spec/acceptance/types/mysql_plugin_spec.rb
hunner: The `shell()` command will actually call out to a `sh`, so you would have to call `shell('mysql -e "select plugin_name ...')` or something to get the mysql output.

dveeden: Good catch. I've updated that line.

dveeden:
I've added a section to the README

igalic:
thaaaaaaaaanks @dveeden!

PR #642, opened 2015-01-20
remove dependency on mysql_user from /root/.my.cnf

matthewfischer:
You cannot create a mysql_user until the .my.cnf file is written so this
dependency does not make sense.

Closes: MODULES-1058


dveeden:
Would it be possible to combine this with htttps://tickets.puppetlabs.com/browse/MODULES-1550 ?

matthewfischer:
Although this still seems wrong to me, making the change is causing other issues that I've found now after more testing, so I'm dropping this.

PR #643, opened 2015-01-20
Make mysql::db more flexible

stamak:
Allow to use the mysql::db if Class['mysql::server'] is not defined
and mysql server is installed by hands or by using another approaches.

Closes-bug: #MODULES-1685


mhaskel:
@stamak thanks for the contribution, but it looks like this is implementing the same functionality as #586 so we're going to close this for now. That PR still needs a little bit of work, so if you want to help get that one finished up that would be great!

PR #644, opened 2015-01-22
Revert "Support for authentication plugins"

cmurphy:
Reverts puppetlabs/puppetlabs-mysql#640

Authentication plugins are not supported on older versions of MySQL [1], which are installed by default on most supported operating systems. We would like to come up with a better way of supporting this feature without breaking the providers for older version of MySQL. In the mean time, we're reverting this pull request while we work out an alternate implementation.

[1] http://dev.mysql.com/doc/refman/5.5/en/pluggable-authentication.html

### DIFF
--- a/lib/puppet/provider/mysql_user/mysql.rb
+++ b/lib/puppet/provider/mysql_user/mysql.rb
@@ -12,14 +12,13 @@ Puppet::Type.type(:mysql_user).provide(:mysql, :parent => Puppet::Provider::Mysq
     # To reduce the number of calls to MySQL we collect all the properties in
     # one big swoop.
     users.collect do |name|
-      query = "SELECT MAX_USER_CONNECTIONS, MAX_CONNECTIONS, MAX_QUESTIONS, MAX_UPDATES, PASSWORD, PLUGIN FROM mysql.user WHERE CONCAT(user, '@', host) = '#{name}'"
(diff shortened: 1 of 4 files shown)


cmurphy:
@dveeden we discovered that #640 broke our acceptance tests, we think because authentication plugins aren't supported in the default mysql version. Example:

```
Error: Could not prefetch mysql_user provider 'mysql': Execution of '/usr/bin/mysql -NBe SELECT MAX_USER_CONNECTIONS, MAX_CONNECTIONS, MAX_QUESTIONS, MAX_UPDATES, PASSWORD, PLUGIN FROM mysql.user WHERE CONCAT(user, '@', host) = 'root@127.0.0.1'' returned 1: ERROR 1054 (42S22) at line 1: Unknown column 'PLUGIN' in 'field list'
```
We'd love to see another pull request for this where some of the implementation details for supporting different versions are worked out.

dveeden:
I've updated the code to include version specific code:
Original:
```SQL
SELECT MAX_USER_CONNECTIONS, MAX_CONNECTIONS, MAX_QUESTIONS, 
MAX_UPDATES, PASSWORD, PLUGIN 
FROM mysql.user WHERE CONCAT(user, '@', host) = 'root@127.0.0.1'
```
New:
```SQL
SELECT MAX_USER_CONNECTIONS, MAX_CONNECTIONS, MAX_QUESTIONS, 
MAX_UPDATES, PASSWORD /*!50508 , PLUGIN */ 
FROM mysql.user WHERE CONCAT(user, '@', host) = 'root@127.0.0.1'
```

This statement works on all MySQL version, but will only return the PLUGIN column for the versions (5.5.7+) which support plugins.

https://dev.mysql.com/doc/refman/5.7/en/comments.html

PR #645, opened 2015-01-24
Support authentication plugins

dveeden:
This uses CREATE USER xxx IDENTIFIED WITH yyy

For tests:
 unix_socket is not loaded by default, so this might require:
 install plugin unix_socket soname 'auth_socket.so';

 The mysql_native_password plugin is available by default and
 allows you to also set a password.

Try to make it compatible with MySQL < 5.5.7 it uses version
specific code with "/*!50508 stmt */"

Previous pull request: puppetlabs/puppetlabs-mysql#640 and rollback: puppetlabs/puppetlabs-mysql/pull/644

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,14 @@ mysql_user { 'root@127.0.0.1':
 }
 ```
 
+It is also possible to specify an authentication plugin.
+```
+mysql_user{ 'myuser'@'localhost':
+  ensure                   => 'present',
+  plugin                   => 'unix_socket',
+}
+```
+
 ####mysql_grant
 
 `mysql_grant` can be used to create grant permissions to access databases within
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -539,6 +539,14 @@ mysql_user { 'root@127.0.0.1':
 }
 ```
 
+It is also possible to specify an authentication plugin.
+```
+mysql_user{ 'myuser'@'localhost':
+  ensure                   => 'present',
+  plugin                   => 'unix_socket',
```
igalic: this alignment here is incorrect. we only need one space there.

hunner:
I'm running this through our acceptance tests before merge this time :)

igalic:
:+1: @hunner :D

PR #647, opened 2015-01-27
Support for more backup methods

dveeden:
Add support for logical backups with:
 * MySQL Enteprise Backup
 * Percona XtraBackup

Possible Issues:
* There is no apt/deb for the meb package: https://bugs.mysql.com/bug.php?id=75633&thanks=sub
* The meb rpm must be downloaded for Oracle https://edelivery.oracle.com https://support.oracle.com

### DIFF
--- a/manifests/server/backup.pp
+++ b/manifests/server/backup.pp
@@ -3,6 +3,7 @@ class mysql::server::backup (
   $backupuser,
   $backuppassword,
   $backupdir,
+  $backupmethod = 'mysqldump',
   $backupdirmode = '0700',
   $backupdirowner = 'root',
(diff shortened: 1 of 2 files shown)


igalic:
not a fan of a giant `case` statement.
i'd much rather have "providers", and we just defer to those.
the only issue with that is, that we'd have to again, either:
* use a `case` statement to call them, or
* use a `create_resources()`

the latter requires that all "providers" have the [same parameters](https://github.com/Brainsware/puppet-trafficserver/blob/types/manifests/storage.pp#L38-L46)

hunner:
Closed in favor of #649 style.

PR #649, opened 2015-01-27
Use backup providers

dveeden:
Add backup providers for
* mysqldump
* MySQL Enterprise Backup
* Percona XtraBackup

### DIFF
--- a/manifests/backup/mysqlbackup.pp
+++ b/manifests/backup/mysqlbackup.pp
@@ -1,0 +1,103 @@
+# See README.me for usage.
+class mysql::backup::mysqlbackup (
+  $backupuser,
+  $backuppassword,
+  $backupdir,
+  $backupdirmode = '0700',
+  $backupdirowner = 'root',
+  $backupdirgroup = 'root',
+  $backupcompress = true,
+  $backuprotate = 30,
+  $ignore_events = true,
+  $delete_before_dump = false,
+  $backupdatabases = [],
+  $file_per_database = false,
+  $ensure = 'present',
+  $time = ['23', '5'],
+  $postscript = false,
+  $execpath   = '/usr/bin:/usr/sbin:/bin:/sbin',
+) {
+
+  mysql_user { "${backupuser}@localhost":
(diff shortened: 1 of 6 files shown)


dveeden:
This is an alternative for #647

igalic:
i'm kinda surprised this is still passing spec tests, for one, for the other, as you can imagine, i prefer this solution, and would as such prefer it to be (better?) documented

manifests/server/backup.pp
```diff
@@ -1,62 +1,43 @@
...
-  }
-
-  file { 'mysqlbackupdir':
-    ensure => 'directory',
-    path   => $backupdir,
-    mode   => $backupdirmode,
-    owner  => $backupdirowner,
-    group  => $backupdirgroup,
-  }
+  create_resources("mysql::backup::${provider}", {
```
hunner: Could you convert the backup defines to classes (since they will only ever be used once) and have this create_resources of type "class")?

hunner:
Descriptions in the readme about the different providers would be great, and document the limitations/requirements (eg: the `meb` package needs to exist in a repo).

manifests/backup/mysqlbackup.pp
```diff
@@ -0,0 +1,85 @@
+# See README.me for usage.
+define mysql::backup::mysqlbackup (
```
hunner: Should this be called mysqlenterprisebackup instead?

dveeden: No, I don't think so. The full name is MySQL Enterprise Backup and the
package name is meb and the tool/binary which you actually need to call
is mysqlbackup.

manifests/server/backup.pp
dveeden: @hunner done. Looks better.

dveeden:
I've added a short description to the README.md and I've added a config file for mysqlbackup (separate from the main my.cnf so it can be protected with 0600).

igalic:
thaaaaanks @dveeden!
and thank you @hunner for the review!

PR #651, opened 2015-01-27
(#MODULES-1058) root_password.pp cannot create /root/.my.cnf due to depe...

lodgenbd:
...ndency

The dependency of creating the root .my.cnf file is a command which requires
the .my.cnf file. This patch removes that dependency. Without removing the
dependency, if a user already has a mysql server installed with a root password
and no root .my.cnf file, the module application will fail.

### DIFF
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -14,7 +14,6 @@ class mysql::server::root_password {
       content => template('mysql/my.cnf.pass.erb'),
       owner   => 'root',
       mode    => '0600',
-      require => Mysql_user['root@localhost'],
     }
   }
 


lodgenbd:
Further details available at https://tickets.puppetlabs.com/browse/MODULES-1058

matthewfischer:
I tried similar but it seemed to cause other issues in my environment. I ran out of time to dive in more, but this does seem like the obvious fix. I will try it again hopefully this week.

igalic:
@lodgenbd help me understand the use-case here:

i have a mysql-server installed, with a *known* root password, and i want to use puppetlabs-mysql to manage it?

how will this change affect the usecase of, i want to install a new mysql server?

lodgenbd:
@igalic I tested my code change in the use case of "i want to install a new mysql server" with no issues.

igalic:
okay, thank you for the clarification, @lodgenbd, and for the patch!

hogepodge:
igalic, fwiw, this patch broke my mysql use case.

lodgenbd:
@hogepodge what's your use case?

hogepodge:
I'm running the puppet-openstack module set. When I run the code from those modules to set up a variety of users and accounts the .my.cnf file is not handled properly, and I get locked out of mysql administration.

lodgenbd:
@hogepodge when you say the file isn't handled properly, what exactly is happening with the file?

lodgenbd:
@cyberious can you share what you're seeing?

hogepodge:
In a while I can post the error I'm seeing. I actually depend on this module for day to day work, so when it broke in my environment I pinned back to a stable version so I could stay productive. What I do know is that previously I could create new users and databases, and my ability to do that with the module broke. I'm not sure exactly what's going on, but if I recall correctly it's telling me that I can't log in as the root user.

lodgenbd:
@cyberious can you confirm if the .my.cnf file was created before that error occurred?

lodgenbd:
It sounds like we need an order relationship so that the .my.cnf file is created before anything that would run a mysql command requiring it. I'm going to assume that's what's happening, unless someone reports that their .my.cnf file is being incorrectly built, which would be another issue altogether.

lodgenbd:
@igalic, Let me know what you think, but I see a couple of issues:

The root user and the .my.cnf file are in the same class, so when the below is called, it tries to take actions before making the .my.cnf file, and they fail. This is in the server.pp:

  } else {
    Anchor['mysql::server::start'] ->
    Class['mysql::server::install'] ->
    Class['mysql::server::config'] ->
    Class['mysql::server::service'] ->
    Class['mysql::server::root_password'] ->
    Class['mysql::server::providers'] ->
    Anchor['mysql::server::end']

If you move the root_password class higher in the order above the install class, it would also fail because the class would try to do two things:

1.) create the root user within mysql
2.) create the .my.cnf file

...before mysql is installed. In that case, #1 above would fail because mysql isn't installed, but #2 would succeed, because it's just a file and doesn't rely on anything.

I believe that #2 should be pulled out and added to a class of its own, so it can be added prior to the "install" class being called, because that class runs mysql commands requiring the .my.cnf file to exist.

If you agree, either one of us can make the changes. I will try to do it tonight if I hear back from you, otherwise I have a training class all day tomorrow and would have to pick it up after that.

cyberious:
@lodgenbd @igalic for now we are reverting the change until a new one that address both use cases can be found.

lodgenbd:
Agreed, thank you

PR #652, opened 2015-01-28
there is no class called mysql anymore (to test)

casharma:
Remove tests/init.pp which seeks to test class mysql.
There is no definition for it, and so there is no need for the init.pp file in tests.


igalic:
that was an easy one, thanks @casharma!

PR #653, opened 2015-01-28
fix config requirement

CharlieR-o-o-t:
used variables at service.pp is not at visible scope

config requirements should absent in cases if config file unmanaged.

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -1,26 +1,39 @@
 #
-class mysql::server::service {
+class mysql::server::service (
+  $real_service_manage  = $mysql::server::real_service_manage,
+  $manage_config_file   = $mysql::server::manage_config_file,
+  $real_service_enabled = $mysql::server::real_service_enabled,
+  $log_error            = $mysql::params::log_error,
+  $service_name         = $mysql::server::service_name,
+  $service_provider     = $mysql::server::service_provider
+) inherits mysql::params {
 
-  if $mysql::server::real_service_manage {
-    if $mysql::server::real_service_enabled {
+  if $real_service_manage {
+    if $real_service_enabled {
       $service_ensure = 'running'
     } else {
       $service_ensure = 'stopped'
     }
   }
 
-  file { $mysql::params::log_error:
+  file { $log_error:
     ensure => present,
     owner  => 'mysql',
     group  => 'mysql',
   }
 
+  if $manage_config_file {
+    $service_require =  [ File['mysql-config-file'], Package['mysql-server'] ]
+  } else {
+    $service_require =  [ Package['mysql-server'] ]
+  }
+
   service { 'mysqld':
     ensure   => $service_ensure,
-    name     => $mysql::server::service_name,
-    enable   => $mysql::server::real_service_enabled,
-    provider => $mysql::server::service_provider,
-    require  => [ File['mysql-config-file'], Package['mysql-server'] ]
+    name     => $service_name,
+    enable   => $real_service_enabled,
+    provider => $service_provider,
+    require  => $service_require 
   }
 
 }


supernovae:
I second this patch, one can't install mysql and not manage the config right now as it complaints..

  class { 'mysql::server':
    manage_config_file => false

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: Service[mysqld] { require => File[mysql-config-file] }, because File[mysql-config-file] doesn't seem to be in the catalog

mhaskel:
@CharlieR-o-o-t I think this can be fixed with a much smaller changeset. If you can simplify the pull request to only include those changes (will make comments inline) I think this should be ok.

manifests/server/service.pp
```diff
@@ -1,26 +1,39 @@
...
     owner  => 'mysql',
     group  => 'mysql',
   }
 
+  if $manage_config_file {
+    $service_require =  [ File['mysql-config-file'], Package['mysql-server'] ]
+  } else {
+    $service_require =  [ Package['mysql-server'] ]
+  }
+
```
mhaskel: Keep this block of changes.

manifests/server/service.pp
```diff
@@ -1,26 +1,39 @@
...
   service { 'mysqld':
     ensure   => $service_ensure,
-    name     => $mysql::server::service_name,
-    enable   => $mysql::server::real_service_enabled,
-    provider => $mysql::server::service_provider,
-    require  => [ File['mysql-config-file'], Package['mysql-server'] ]
+    name     => $service_name,
+    enable   => $real_service_enabled,
+    provider => $service_provider,
+    require  => $service_require
```
mhaskel: Keep the last line of changes (except replace the trailing whitespace with a trailing comma please)

manifests/server/service.pp
```diff
@@ -1,26 +1,39 @@
 #
-class mysql::server::service {
+class mysql::server::service (
+  $real_service_manage  = $mysql::server::real_service_manage,
+  $manage_config_file   = $mysql::server::manage_config_file,
+  $real_service_enabled = $mysql::server::real_service_enabled,
+  $log_error            = $mysql::params::log_error,
+  $service_name         = $mysql::server::service_name,
+  $service_provider     = $mysql::server::service_provider
+) inherits mysql::params {
```
mhaskel: Revert these changes.

manifests/server/service.pp
```diff
@@ -1,26 +1,39 @@
...
 
-  if $mysql::server::real_service_manage {
-    if $mysql::server::real_service_enabled {
+  if $real_service_manage {
+    if $real_service_enabled {
```
mhaskel: Revert these changes

manifests/server/service.pp
```diff
@@ -1,26 +1,39 @@
...
       $service_ensure = 'running'
     } else {
       $service_ensure = 'stopped'
     }
   }
 
-  file { $mysql::params::log_error:
+  file { $log_error:
```
mhaskel: Revert this change

manifests/server/service.pp
```diff
@@ -1,26 +1,39 @@
...
+
   service { 'mysqld':
     ensure   => $service_ensure,
-    name     => $mysql::server::service_name,
-    enable   => $mysql::server::real_service_enabled,
-    provider => $mysql::server::service_provider,
-    require  => [ File['mysql-config-file'], Package['mysql-server'] ]
+    name     => $service_name,
+    enable   => $real_service_enabled,
+    provider => $service_provider,
```
mhaskel: Revert changes up to here.

igalic:
thank you everyone! this was fixed in #672

PR #654, opened 2015-01-29
Check for full path for log-bin to stop puppet from managing directory “."

NoodlesNZ:
As reported by @larkit-ian in PR #596

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -22,12 +22,16 @@ class mysql::server::config {
   $logbin = pick($options['mysqld']['log-bin'], $options['mysqld']['log_bin'], false)
 
   if $logbin {
-    $logbindir = mysql_dirname($logbin)
-    file { $logbindir:
-      ensure => directory,
-      mode   => '0755',
-      owner  => $options['mysqld']['user'],
-      group  => $options['mysqld']['user'],
+    $logbindir = dirname($logbin)
+    
+    #Stop puppet from managing directory if just a filename/prefix is specified
+    if $logbindir != '.' {
+      file { $logbindir:
+        ensure => directory,
+        mode   => '0755',
+        owner  => $options['mysqld']['user'],
+        group  => $options['mysqld']['user'],
+      }
     }
   }
 


manifests/server/config.pp
```diff
@@ -23,11 +23,15 @@
...
   if $logbin {
     $logbindir = dirname($logbin)
-    file { $logbindir:
-      ensure => directory,
-      mode   => '0755',
-      owner  => $options['mysqld']['user'],
-      group  => $options['mysqld']['user'],
+    
+    #Stop puppet from managing directory if just a filename/prefix is specified
+    if $logbindir != '.' {
```
igalic: i'm confused about this: when do we (anyone?) set logbindir to .? how do we know to set it to .?

larkit-ian:
As I read it, if we don't specify an absolute path, then dirname($logbin) will set $logbindir to ".".  In that case, we don't to have the file resource for that dir.

JCotton1123:
I am also currently running into this issue. We can please get this merged ASAP?

@NoodlesNZ it looks like you need to rebase your branch

NoodlesNZ:
I have rebased, but CI is taking a while.

igalic:
i see now how this makes sense ~_~

NoodlesNZ:
:) Thanks for the merge

manifests/server/config.pp
```diff
@@ -22,12 +22,16 @@
   $logbin = pick($options['mysqld']['log-bin'], $options['mysqld']['log_bin'], false)
 
   if $logbin {
-    $logbindir = mysql_dirname($logbin)
-    file { $logbindir:
-      ensure => directory,
-      mode   => '0755',
-      owner  => $options['mysqld']['user'],
-      group  => $options['mysqld']['user'],
+    $logbindir = dirname($logbin)
```
mhaskel: This needs to use `mysql_dirname` not `dirname` as the stdlib requirement is > 3.2.0.

mhaskel: Issued #677 to fix this.

PR #655, opened 2015-01-30
(MODULES-1731) Invalid parameter 'provider' removed from mysql_user instance.

rnelson0:


hunner:
Odd, that should work as the provider is actually called `mysql`. Though since there is only one provider, it's not required. https://github.com/puppetlabs/puppetlabs-mysql/blob/master/lib/puppet/provider/mysql_user/mysql.rb#L2

rnelson0:
That's what I thought, but I could not determine why the provider fails every time.

optiz0r:
This issue has been reintroduced during the refactor for 3.3.0

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter provider on Mysql_user[backups@localhost] at /home/puppet/environments/dev/thirdparty/mysql/manifests/backup/mysqldump.pp:26 on node sql1.local     
Warning: Not using cache on failed catalog
```

igalic:
i'm confused. is this failing on master for you?

this seems to be failing for different reasons…

optiz0r:
Hi @igalic,

In 3.1 mysql::server::backup explicitly set the provider parameter on mysql_user["${backupuser}@localhost"]. That didn't work and prompted this PR in which the offending line was removed. Between 3.2 and 3.3 the mysql::server::backup class was refactored into mysql::backup::mysqldump which re-added the previously removed provider. I'm now experiencing the same error message that this PR originally solved from the new location (as shown in my previous comment).

Comparing the 3.3.0 tree with master I suspect master would still be affected. This looks like a regression.

I've commented on this thread because it had the right context, but perhaps opening a new issue against 3.3.0 would be more appropriate?

igalic:
@optiz0r please see #687.

PR #659, opened 2015-02-06
Update version dependency on stdlib

andersbs:
mysql::server::config uses the dirname function from stdlib, which appeared in 4.1.0.

This isn't a problem if you install from the forge but my manual validation led me to belive I didn't have to update my existing stdlib.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -76,7 +76,7 @@
   ],
   "description": "Mysql module",
   "dependencies": [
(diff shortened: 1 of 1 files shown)


igalic:
@andersbs do you wanna add your commit email to your github settings?
****
i really hope we're not running into any conflicts with this :S

hunner:
Thanks for catching this!

PR #660, opened 2015-02-06
Bugfix: increase minimum stdlib

hunner:
Mysql 3.1.0 included 08a66b7 which uses dirname() which was released in
stdlib 4.1.0. This change bumps the minimum stdlib up to 4.1.0

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -76,7 +76,7 @@
   ],
   "description": "Mysql module",
   "dependencies": [
-    {"name":"puppetlabs/stdlib","version_requirement":">= 3.2.0"},
+    {"name":"puppetlabs/stdlib","version_requirement":">= 4.1.0 < 5.0.0"},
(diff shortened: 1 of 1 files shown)


hunner:
Oh just noticed #659. This PR also adds the upper bound of 5.0.0

metadata.json
```diff
@@ -76,7 +76,7 @@
   ],
   "description": "Mysql module",
   "dependencies": [
-    {"name":"puppetlabs/stdlib","version_requirement":">= 3.2.0"},
+    {"name":"puppetlabs/stdlib","version_requirement":">= 4.1.0 < 5.0.0"},
```
cyberious: Do we need to bump the lower range to above 4.1?  if so can't we just use ~>4.1

hunner: I don't think `~>` works for modules, just gems.

cyberious: yeah it would be 4.x and that won't cover it.

PR #671, opened 2015-02-12
Fix bug in 578

aldavud:
the recently added feature to support galera by allowing independent
creation of the root@localhost user in the DB and the /root/.my.cnf
file contains a bug.

specifically the .my.cnf file resource still requires the root@localhost
resource, even when it is not available.

this fixes the issue by making the dependency optional.

### DIFF
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -16,7 +16,9 @@ class mysql::server::root_password {
       content => template('mysql/my.cnf.pass.erb'),
       owner   => 'root',
       mode    => '0600',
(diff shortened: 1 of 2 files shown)


igalic:
@aldavud would you very much mind [squashing](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) those two commits?

aldavud:
There you go...

igalic:
thanks @aldavud!

aldavud:
thank you!

PR #678, opened 2015-02-27
(MODULES-1804) Allow override of log-error

hunner:
The mysqld log-error setting may be passed by override_options, but
server/service.pp was not using the overridden value. This fixes that.

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -1,5 +1,6 @@
 #
 class mysql::server::service {
+  $options = $mysql::server::options
 
   if $mysql::server::real_service_manage {
     if $mysql::server::real_service_enabled {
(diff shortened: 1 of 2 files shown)


manifests/server/service.pp
```diff
@@ -9,7 +10,7 @@
     }
   }
 
-  file { $mysql::params::log_error:
```
mhaskel: will log-error always be specified?

hunner: I believe so, yes. Otherwise we would have seen compile errors because of undef titles :). I could surround it with an `if` but I don't know if not having an error log is a valid use case.

mhaskel:
@hunner works for me!

PR #682, opened 2015-03-03
Remove default install root password if set

JCotton1123:
I haven't written any tests for this PR b/c I wasn't sure you'd be willing to merge it. If you're OK with this approach, I will write tests. If you're not, please provide some feedback on an alternate way to address this "feature".

New installations of MySQL will configure a default random password for the root user with an expiration. See http://datacharmer.blogspot.com/2012/11/mysql-568-broken-compatibility-ahead.html. Until the root password is changed, no actions can be performed. This PR adds detection of the mysql_secret file and removes this password. If the user desires to set a password it will be configured by the mysql_user resource further down in the mysql::server::root_password class.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -6,6 +6,7 @@ class mysql::params {
   $purge_conf_dir         = false
   $restart                = false
   $root_password          = 'UNSET'
+  $install_secret_file    = '/.mysql_secret'
   $server_package_ensure  = 'present'
   $server_package_manage  = true
   $server_service_manage  = true
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -2,12 +2,28 @@
 class mysql::server::root_password {
 
   $options = $mysql::server::options
+  $secret_file = $mysql::server::install_secret_file
+
+  # New installations of MySQL will configure a default random password for the root user
+  # with an expiration. No actions can be performed until this password is changed. The
+  # below exec will remove this default password. If the user has supplied a root
+  # password it will be set further down with the mysql_user resource.
+  $rm_pass_cmd = join([
+    "mysqladmin -u root --password=\$(grep -o '[^ ]\\+\$' ${secret_file}) password ''",
+    "rm -f ${secret_file}"
+  ], ' && ')
+  exec { 'remove install pass':
+    command => $rm_pass_cmd,
+    onlyif  => "test -f ${secret_file}",
+    path    => '/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin'
+  }
 
   # manage root password if it is set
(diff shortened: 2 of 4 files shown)


igalic:
i'm really unclear about the mechanics of this.

manifests/params.pp
```diff
@@ -6,6 +6,7 @@
   $purge_conf_dir         = false
   $restart                = false
   $root_password          = 'UNSET'
+  $install_secret_file    = '/.mysql_secret'
```
igalic: is this supposed to be `/`, rather than `/root/` ?

manifests/server/root_password.pp
```diff
@@ -2,12 +2,27 @@
...
 
   $options = $mysql::server::options
+  $secret_file = $mysql::server::install_secret_file
+
+  # New installations of MySQL will configure a default random password for the root user
+  # with an expiration. No actions can be performed until this password is changed. The
+  # below exec will remove this default password. If the user has supplied a root
+  # password it will be set further down with the mysql_user resource.
+  $ch_pass_command = join([
+    "/usr/bin/mysqladmin -u root --password=\$(awk '/password/{print \$NF}' ${secret_file}) password ''",
```
igalic: this seems like a really awkward way to get at the password.
i'm not sure /usr/bin/mysqladmin is what we want here. on other systems it might be in different locations, it might be better to work with path => instead

daenney: Yeah, hard coded paths are a bad idea. As a consequence, this really ought to be covered by an acceptance test too so we can catch platform differences.

manifests/params.pp
JCotton1123: When Puppet installs the MySQL-server package this file seems to end up at /.mysql_secret

manifests/server/root_password.pp
JCotton1123: Thanks for the feedback. As I said, I didn't expect this to get merged. This was a quick attempt at addressing the problem. I was half expecting you to ask that I replace the exec with a type/provider.

I will change the expression used to retrieve the password so its clearer, I will remove the hardcoded paths and add the path parameter, and I will add tests to cover this.

manifests/params.pp
igalic: on which platform?

JCotton1123: CentOS, Oracle MySQL

manifests/server/root_password.pp
daenney: Heh. We don't tend to ask people to write Types and Providers until that "mess" is fixed in Puppet. It's just not feasible in most cases and testing them is currently not in a good state either.

JCotton1123:
Hi, this has been hanging around for a while. Can you please let me know if it needs any additional work to make it mergable.

igalic:
@JCotton1123 thanks for the reminder.

PR #683, opened 2015-03-03
Remove references to old_root_password (cleanup)

JCotton1123:
As far as I could tell this parameter is not used anywhere.


JCotton1123:
I can revise this PR, revert the changes, and just update the README file with a note that this parameter exists for compatibility and doesn't do anything. Does that make sense?

JCotton1123:
I've updated the PR and indicated that `old_root_password` is deprecated.

igalic:
this looks betterer, i think we could merge it as it is, though i'm not sure about the semantics of introducing warnings between versions.

JCotton1123:
Hi, this has been hanging around for a while. Can you please let me know if it needs any additional work to make it mergable.

igalic:
@mhaskel: ping

mhaskel:
@JCotton1123 sorry I missed the updates on this, but looks good now, so thanks!

PR #687, opened 2015-03-05
mysql backup: fix regression in mysql_user call

igalic:
due to a mishappen rebase, in #649, we introduced a regression fixed
which was fixed in #655.

how come our tests don't catch this?


mhaskel:
thanks @igalic

PR #690, opened 2015-03-17
Cosmetic: add comment managed by puppet

mremy:
Add a comment in the .my.cnf file, to show that the file is managed by puppet, as in the other templates.


igalic:
thanks @mremy!

PR #691, opened 2015-03-18
Fix regression introduced by adding OpenBSD support.

buzzdeee:
The pull request referring to was #567 

Add a $mysql_group parameter, and use that instead of the $root_group
parameter to define the group membership of the mysql error log file.

For all OS but OpenBSD, the value is set to 'mysql', which was hardcoded
in manifests/server/service.pp before, for OpenBSD, it's set to '_mysql'.

Should bring back old behavior on non-OpenBSD, and some tests showed, 
no problems on OpenBSD.


mhaskel:
\o/ thanks @buzzdeee !

PR #692, opened 2015-03-19
set show_diff to false for file with password

applewiskey:
It's no secure to be able to view passwords in logs or interactively.

### DIFF
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -13,12 +13,18 @@ class mysql::server::root_password {
 
   if $mysql::server::create_root_my_cnf == true and $mysql::server::root_password != 'UNSET' {
     file { "${::root_home}/.my.cnf":
+      ensure  => file,
       content => template('mysql/my.cnf.pass.erb'),
       owner   => 'root',
       mode    => '0600',
     }
     if $mysql::server::create_root_user == true {
       Mysql_user['root@localhost'] -> File["${::root_home}/.my.cnf"]
+    }
+    if versioncmp($::puppetversion, '3.2.0') =~ /[01]/ {
+      File["${::root_home}/.my.cnf"] {
+        show_diff => false,
+      }
     }
   }
 


igalic:
oh the joy… this fails with older puppets: https://travis-ci.org/puppetlabs/puppetlabs-mysql/jobs/55012360#L2225-L2239

petems:
Yeah, testing things that don't exist in older Puppet is a bit tricky, you need to do something like this:

```
if (Puppet.version >= '3.5.0')
    context "show_diff parameter with puppet #{Puppet.version}" do
      let(:facts) {{ :puppetversion => Puppet.version }}
      it { is_expected.to contain_file('/root/.my.cnf') }
  else
    context "validating content with puppet #{Puppet.version}" do
       let(:facts) {{ :puppetversion => Puppet.version }}
       it { is_expected.to contain_file('/root/.my.cnf') }
    end
  end
```

To ensure the fact is set based on the Puppet version.

petems:
Here's a similar test for the validate_cmd parameter which is Puppet > 3.5 only

https://github.com/arnoudj/puppet-sudo/blob/master/spec/defines/sudo_sudoers_spec.rb#L48

igalic:
btw, the reason for the failing tests:
```
 1) mysql::server on PE3.0 centos-6-x86_64 mysql::server::root_password when root_password set should contain Mysql_user[root@localhost]
     Failure/Error: it { is_expected.to contain_mysql_user('root@localhost') }
     Puppet::Error:
       Evaluation Error: Left match operand must result in a String value. Got an Integer. at /home/travis/build/puppetlabs/puppetlabs-mysql/spec/fixtures/modules/mysql/manifests/server/root_password.pp:24:8 on node testing-worker-linux-docker-73fba05c-3373-linux-16.prod.travis-ci.org
```

jonnytpuppet:
Hi There

I'm closing this PR due to inactivity. It would be great if you could fix the raised concerns and re-open another PR, as I'd love to see this merged!

Thanks!

PR #693, opened 2015-03-23
Change instructions for running tests to use 'bundle exec'.

spil-jaak:
There is no ticket for this. I'll create one, if it is necessary.


igalic:
can you please submit that fix to the upstream document?
https://github.com/puppetlabs/modulesync_configs/blob/master/moduleroot/CONTRIBUTING.md

PR #697, opened 2015-03-30
fix FreeBSD support for backups

fraenki:
The `mysql::backup::*` classes still use the hardcoded group name `root`, which is not available on FreeBSD systems. This patch changes this to use the param `mysql::params::root_group` instead. Besides that it utilizes `/bin/sh` instead of `/bin/bash` on systems other than Linux.

### DIFF
--- a/manifests/backup/mysqlbackup.pp
+++ b/manifests/backup/mysqlbackup.pp
@@ -5,7 +5,7 @@ class mysql::backup::mysqlbackup (
   $backupdir,
   $backupdirmode = '0700',
   $backupdirowner = 'root',
-  $backupdirgroup = 'root',
+  $backupdirgroup = $mysql::params::root_group,
   $backupcompress = true,
   $backuprotate = 30,
   $ignore_events = true,
--- a/templates/mysqlbackup.sh.erb
+++ b/templates/mysqlbackup.sh.erb
@@ -1,4 +1,8 @@
+<%- if @kernel == 'Linux' -%>
 #!/bin/bash
+<%- else -%>
+#!/bin/sh
+<%- end -%>
 #
 # MySQL Backup Script
 #  Dumps mysql databases to a file for another backup tool to pick up.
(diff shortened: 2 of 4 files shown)


manifests/backup/mysqlbackup.pp
```diff
@@ -5,7 +5,7 @@
   $backupdir,
   $backupdirmode = '0700',
   $backupdirowner = 'root',
-  $backupdirgroup = 'root',
+  $backupdirgroup = $mysql::params::root_group,
```
igalic: shouldn't this be $mysql::root_group?

templates/mysqlbackup.sh.erb
```diff
@@ -1,4 +1,8 @@
+<%- if @kernel == 'Linux' -%>
 #!/bin/bash
```
igalic: we can probably set this unconditionally, since we're not really using any oh…

templates/mysqlbackup.sh.erb
```diff
@@ -27,7 +31,9 @@ PATH=<%= @execpath %>
 
 
 
+<%- if @kernel == 'Linux' -%>
 set -o pipefail
```
igalic: we actually are…

igalic: wondering if we can make the shell dependent on whether `/bin/bash` actually exists… but then again, FreeBSD puts bash in `/usr/local/bin` (and doesn't install it per default)

manifests/backup/mysqlbackup.pp
fraenki: Well, we *might* use `$mysql::root:group` here, but this module's style is to use `$mysql::params::*` elsewhere.

templates/mysqlbackup.sh.erb
fraenki: If simplification would be the *main* goal, I'd always opt to change the script to rely on `/bin/sh` instead of bash for obvious compatibility reasons. Otherwise I'm fine with my suggested fallback on non-Linux platforms...

igalic:
true, true!
thanks @

PR #698, opened 2015-04-13
MODULES-1928 - allow log-error to be undef

mhaskel:
Wrap the file resource so that this doesn't explode if log-error is
undef. This is required for syslogging.

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -18,10 +18,12 @@ class mysql::server::service {
     $mysqluser = $options['mysqld']['user']
   }
 
-  file { $options['mysqld']['log-error']:
-    ensure => present,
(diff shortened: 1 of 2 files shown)


manifests/server/service.pp
```diff
@@ -18,10 +18,12 @@
     $mysqluser = $options['mysqld']['user']
   }
 
-  file { $options['mysqld']['log-error']:
-    ensure => present,
-    owner  => $mysqluser,
-    group  => $::mysql::server::mysql_group,
+  if $options['mysqld']['log-error'] {
```
igalic: will this work in strict mode?

mhaskel: yep, works in strict mode

PR #699, opened 2015-04-15
Adds default values for parameters and align assignments

melan:
- Adds default values to parameters in few files
- Adds a references to README.md at the beginning of couple files
- Adds parameters vertical alignment

Result: less puppet-lint warnings


igalic:
thanks!

PR #701, opened 2015-04-23
Split package installation and database install

roidelapluie:
The current install pattern used by this module is the following:
1. Install package
2. Install the DB
3. Make the config files

In some cases this prevents MySQL to start, because some variables in
the config file have an impact on the DB installation.

Example with a custom innodb_data_file_path:
[ERROR] InnoDB: space header page consists of zero bytes in data file
/var/lib/mysql/ibdata

This commit changes the order to do:
1. Install package
2. Make the config files
3. Install the DB

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -51,6 +51,7 @@ class mysql::server (
 
   include '::mysql::server::install'
   include '::mysql::server::config'
+  include '::mysql::server::installdb'
   include '::mysql::server::service'
   include '::mysql::server::root_password'
   include '::mysql::server::providers'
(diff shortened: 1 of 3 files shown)


roidelapluie:
This PR does not bring any API change.

manifests/server.pp
```diff
@@ -65,24 +66,18 @@
...
-    Anchor['mysql::server::end']
-  } else {
-    Anchor['mysql::server::start'] ->
-    Class['mysql::server::install'] ->
-    Class['mysql::server::config'] ->
-    Class['mysql::server::service'] ->
-    Class['mysql::server::root_password'] ->
-    Class['mysql::server::providers'] ->
-    Anchor['mysql::server::end']
+    Class['mysql::server::service']
```
igalic: this strikes me as confusing, especially since we also have

manifests/server.pp
```diff
@@ -65,24 +66,18 @@
...
   }
 
+  Anchor['mysql::server::start'] ->
+  Class['mysql::server::install'] ->
+  Class['mysql::server::config'] ->
+  Class['mysql::server::installdb'] ->
+  Class['mysql::server::service'] ->
```
igalic: service here.

manifests/server.pp
roidelapluie: yes but this is a notify triggered only if $restart is true

manifests/server.pp
roidelapluie: do not look at the diff look at the plain file it is more clear. in the first case it is a ~> that happens only of $restart

igalic: ack!

hikaru-noda:
Hi.
The Ubuntu (I use 14.04) runs dpkg-reconfigure mysql-server-5.x after package installation. After that mysql::server module writes configuration file and if you change the innodb_data_file_path, you can't start mysql server anymore.

I think for Ubuntu installation order should be like this one:
1. Make the config files
2. Install package
3. Install the DB

hikaru-noda:
Please take a look at my changes here:
[Added creation of mysql configuration directory for Debian systems.](https://github.com/hikaru-noda/puppetlabs-mysql/commit/17a3a0dab9eba9aca5504d2bd22f2c7136c2072c)

PR #702, opened 2015-04-26
Added $onlyif_empty to db.pp to facilitate once-only import of sql datab...

gibbsoft:
...ase file, rather than import at every puppet run.

I modified db.pp to conditionally check if there are any existing tables in the database, if the schema already has tables then we may not wish to import the specified sql flat file.

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -70,7 +70,7 @@ define mysql::db (
         timeout     => $import_timeout,
         onlyif      => $onlyif_empty ? {
           true    => "test $(mysql ${dbname} -e 'show tables' |wc -l ) -eq 0",
-          default => "false",
+          default => "test $(false)",
         }
       }
     }


manifests/db.pp
```diff
@@ -67,6 +68,10 @@
         require     => Mysql_grant["${user}@${host}/${table}"],
         subscribe   => Mysql_database[$dbname],
         timeout     => $import_timeout,
+        onlyif      => $onlyif_empty ? {
+          true    => "mysql ${dbname} -e 'show tables' |wc -l | expr 1 /",
+          default => "true",
+        }
```
igalic: this shouldn't be quoted

gibbsoft:
I realised that I'd made a mistake, so rolled out my commit and replaced it with another.  The onlyif command now uses 'test'.  The default of "true" is intentionally quoted as it is a call to /bin/true to derive a return value of 0.

manifests/db.pp
```diff
@@ -67,6 +68,10 @@
         require     => Mysql_grant["${user}@${host}/${table}"],
         subscribe   => Mysql_database[$dbname],
         timeout     => $import_timeout,
+        onlyif      => $onlyif_empty ? {
+          true    => "test $(mysql ${dbname} -e 'show tables' |wc -l ) -eq 0",
+          default => "false",
```
igalic: the reason tests fail on Travis is this line. you're "quoting" a boolean value.

gibbsoft: It's a false positive: I'm not really quoting a boolean value - it's a call to /bin/false.

gibbsoft: I've wrapped in in 'test' instead, I'll close this pull-request and raise another against the new commit.

PR #703, opened 2015-04-27
MODULES-1981: Revoke and grant difference of old and new privileges

pondohva:
We have 2 problems with current privileges change behaviour with puppetlabs-mysql:
1. If user doesn't have GRANT privilege, puppet will fail
2. if we want to add ALTER privilege (for example) to ['SELECT', 'INSERT'], we don't want to REVOKE SELECT, INSERT before.

### DIFF
--- a/lib/puppet/provider/mysql_grant/mysql.rb
+++ b/lib/puppet/provider/mysql_grant/mysql.rb
@@ -96,18 +96,20 @@ Puppet::Type.type(:mysql_grant).provide(:mysql, :parent => Puppet::Provider::Mys
     exists? ? (return true) : (return false)
   end
 
-  def revoke(user, table)
+  def revoke(user, table, revoke_privileges = ['ALL'])
     user_string = self.class.cmd_user(user)
     table_string = self.class.cmd_table(table)
-
+    priv_string = self.class.cmd_privs(revoke_privileges)
     # revoke grant option needs to be a extra query, because
     # "REVOKE ALL PRIVILEGES, GRANT OPTION [..]" is only valid mysql syntax
     # if no ON clause is used.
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/mysql_grant/mysql.rb
```diff
@@ -129,11 +131,28 @@ def flush
...
+    diff = diff_privileges(@property_hash[:privileges], privileges)
+    if not diff[:revoke].empty?
+      revoke(@property_hash[:user], @property_hash[:table], diff[:revoke])
+    end
+    if not diff[:grant].empty?
+      grant(@property_hash[:user], @property_hash[:table], diff[:grant], @property_hash[:options])
+    end
     @property_hash[:privileges] = privileges
-
+    mysql([defaults_file, '-NBe', 'FLUSH PRIVILEGES'].compact)
```
hunner: Flush privileges is already being done in the `flush` method, so we don't need it here.

pondohva: Yes, there is flush method, but I can't find place in module, where this method is called.

DavidS: Flush is automatically called by the puppet transaction after the resource
is finished.

pondohva:
I've created new pull request https://github.com/puppetlabs/puppetlabs-mysql/pull/706, 'cause this one was deleted with branch (maybe I should create new commit instead git commit -a --amend).

Thank you, @DavidS, can you tell me where I can find this flush call?

DavidS:
`git commit --amend` is OK, you need to `git push -f`instead of delete/push though, to keep github happy.

The flush is done here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type.rb#L985

PR #704, opened 2015-04-27
Added $onlyif_empty to db.pp

gibbsoft:
Added $onlyif_empty to db.pp to facilitate once-only import of sql database file, rather than import at every puppet run.

I modified db.pp to conditionally check if there are any existing tables in the database, if the schema already has tables then we may not wish to import the specified sql flat file.

This commit uses 'test $(false)' rather than the call to 'false' directly to avoid upsetting the CI.


igalic:
@gibbsoft, please take a look at [github flow](https://guides.github.com/introduction/flow/index.html)
that work-flow should make it easier for you to [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing) and [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) commits.

jonnytpuppet:
Hi There

I've been playing around with this PR today trying to get a test created. I had to change some of the logic to get it working (at least in my environment) but I have a few questions:

Was use case do you see this parameter being useful? As it stands now, the module will not import the sql on every Puppet run, unless enforce_sql is set to true. Is this the parameter you were looking for when you had this PR in mind?

Thanks!

gibbsoft:
Hi Jonny,

Yes, I added the code in the pull request because in my environment I found that the database kept getting clobbered.  Although I didn't notice the existing mechanism of using $enforce_sql that you point out, checking back in my calling code, $enforce_sql would have been set to the default of false.

hunner:
@gibbsoft Could you please provide exact conditions (example manifest code, preceding setup steps, etc) to reproduce what you are seeing? And any output you have would be helpful too.

manifests/db.pp
```diff
@@ -57,6 +58,11 @@
 
     $refresh = ! $enforce_sql
 
+    $onlyif_empty_bool = $onlyif_empty ? {
+      true    => "test $(mysql ${dbname} -e 'show tables' |wc -l ) -eq 0",
```
DavidS: `dbname` here requires shellquoting.

DavidS:
Looking at the code, setting `enforce_sql` to false will only set `refreshonly` on the `exec`, which might not be enough in all cases, e.g. when the collaction changes, the sql would be re-run.

gibbsoft:
Thanks David, I meant to try and dig out the usage case where I seen it
work unexpectedly, but so much water has passed under the bridge since then.

Best regards



Nigel
___


Nigel Gibbs
Cloud Architecture Consultant / Linux DevOps Engineer
Gibbsoft Ltd
w: gibbsoft.com
e: nigel@gibbsoft.com
t: +44 (0) 777 197 4018

jonnytpuppet:
@gibbsoft I'm closing this PR as at the present time, we can't find a use-case for this extra param, given we already have $enforce_sql.

Feel free to submit another PR if you feel we're missing a use case.

Thanks!

PR #706, opened 2015-05-04
MODULES-1981: Revoke and grant difference of old and new privileges

pondohva:
We have 2 problems with current privileges change behaviour with puppetlabs-mysql:
1. If user doesn't have GRANT privilege, puppet will fail
2. if we want to add ALTER privilege (for example) to ['SELECT', 'INSERT'], we don't want to REVOKE SELECT, INSERT before.


igalic:
@pondohva one minor issue, before i hit the merge button here: you may wanna change the email on your commit, or add that email to your github settings, so github will know to associate it with your account, and you get magic internet points.

pondohva:
@igalic fixed! Thank you.

PR #707, opened 2015-05-06
Modulesync updates

mhaskel:


igalic:
as i've already said in https://github.com/puppetlabs/modulesync_configs/pull/19 i think this needs puppet 4.0 and strict variables.

PR #708, opened 2015-05-07
(fix) - Check for mysql_verison before assuming that triggers are a valid permission

cyberious:


mhaskel:
Tests and docs should be added for the new fact

lib/facter/mysql_version.rb
```diff
@@ -0,0 +1,8 @@
+Facter.add("mysql_version") do
+  setcode do
+    mysql_ver = Facter::Util::Resolution.exec('mysql --version')
+    if mysql_ver
+      mysql_ver.match(/\d+\.\d+\.\d+/)[0]
```
igalic: this gives us 10.0.19 for mariadb-client-10.0, and 5.6.24 for mysql-client-core-5.6

i wonder what we can do with that now…

igalic: (asking for a friend, #548)

PR #711, opened 2015-05-11
(fix) - Change default for mysql::server::backup to ignore_triggers =…

cyberious:
…> false

- Update spec acceptance test to check for which version is currently installed prior to testing

### DIFF
--- a/spec/acceptance/mysql_backup_spec.rb
+++ b/spec/acceptance/mysql_backup_spec.rb
@@ -1,6 +1,18 @@
 require 'spec_helper_acceptance'
+require 'puppet'
+require 'puppet/util/package'
 
 describe 'mysql::server::backup class' do
+
+  def pre_run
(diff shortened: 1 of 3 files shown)


spec/acceptance/mysql_backup_spec.rb
```diff
@@ -1,6 +1,18 @@
...
 
 describe 'mysql::server::backup class' do
+
+  def pre_run
+    apply_manifest("class { 'mysql::server': root_password => 'password' }", :catch_failures => true)
+    @mysql_version = (on default, 'mysql --version').output.chomp.match(/\d+\.\d+\.\d+/)[0]
+  end
+
+  def version_is_greater_than(version)
+    return Puppet::Util::Package.versioncmp(@mysql_version, version) > 0
```
igalic: can we reuse this with mariadb long user/db names?

PR #713, opened 2015-05-13
Fixed for mariadb

slalomzacharyd:
There were 2 hardcoded service['mysql-server'] that were replaced with the variable, making this module compatible with mariadb on centos7.

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -11,8 +11,7 @@ class mysql::server::installdb {
 
     if $mysql::server::manage_config_file {
       $install_db_args = "--basedir=${basedir} --defaults-extra-file=${config_file} --datadir=${datadir} --user=${mysqluser}"
-    } else {
-      $install_db_args = "--basedir=${basedir} --datadir=${datadir} --user=${mysqluser}"
+
     }
 
     exec { 'mysql_install_db':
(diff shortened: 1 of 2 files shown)


manifests/server/installdb.pp
```diff
@@ -20,7 +19,7 @@
       creates   => "${datadir}/mysql",
       logoutput => on_failure,
       path      => '/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin',
-      require   => Package['mysql-server'],
+      require   => Package[$mysql::params::server_package_name],
```
igalic: this should be overridable in mysql::server, we shouldn't get it from params

manifests/server/service.pp
```diff
@@ -31,7 +31,7 @@
     name     => $mysql::server::service_name,
     enable   => $mysql::server::real_service_enabled,
     provider => $mysql::server::service_provider,
-    require  => Package['mysql-server'],
+    require  => Package[$mysql::params::server_package_name],
```
igalic: same as above.

slalomzacharyd:
Added the changes you requested.

igalic:
could you please [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) these commits?

slalomzacharyd:
I can.  But it's honestly much easier if you just squash on merge with the commit message that makes sense for you instead of my doing a rebase on a branch that will be thrown out after this.

igalic:
okay! thank you @slalomzacharyd

rebased in #718

PR #714, opened 2015-05-14
Generate configuration file before the package installation

hikaru-noda:
Hi.
The Ubuntu (I use 14.04) runs dpkg-reconfigure mysql-server-5.x after package installation. After that mysql::server module writes configuration file and if you change the innodb_data_file_path, you can't start mysql server anymore.

The current install pattern used by this module is the following:
1. Install package
2. Make the config files
3. Install the DB

I think for Debian/Ubuntu installation order should be like this one:
1. Make the config files
2. Install package
3. Install the DB

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -153,6 +153,7 @@ class mysql::params {
       $server_package_name     = 'mysql-server'
 
       $basedir                 = '/usr'
+      $config_dir	       = '/etc/mysql'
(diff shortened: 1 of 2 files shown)


igalic:
tests are also failing…

we should adapt that pattern, defensively, for all platforms.

jonnytpuppet:
Hi There

I'm closing this PR as the tests have failed. Please re-open this PR once the tests work.

Thanks!

PR #716, opened 2015-05-20
dont print root password in diff

tphoney:


igalic:
@tphoney failing tests :(

tphoney:
yup, i noticed. i will get a look at it later. :(

tphoney:
show_diff is not in puppet 2.7, this is why the unit tests are failing. 2.7 support will be dropped with the next major version increase of the module.

tphoney:
@igalic does this feel ok ?

if versioncmp($::puppetversion, '2.7') >= 0 {
  # Foo
}
else {
  # Bar
}

as a stop gap solution until we remove 2.7 support.

.gitignore
```diff
@@ -7,3 +7,4 @@ spec/fixtures/
 coverage/
 .idea/
 *.iml
+log/
```
EmilienM: this should be part of a separated commit.

manifests/server/root_password.pp
```diff
@@ -12,10 +12,20 @@
   }
 
   if $mysql::server::create_root_my_cnf == true and $mysql::server::root_password != 'UNSET' {
-    file { "${::root_home}/.my.cnf":
-      content => template('mysql/my.cnf.pass.erb'),
-      owner   => 'root',
-      mode    => '0600',
+    # Puppet 2.7 doesnt support show_diff
+    if versioncmp($::puppetversion, '3.0') <= 0 {
```
EmilienM: It should be >= 0  to test Puppet 3.0, 3.1, ..., no?

tphoney:
closed in preference of #732

PR #718, opened 2015-05-27
Fixed server package name so it isn't hardcoded to mysql

igalic:
this is a rebase of  #713


igalic:
again, @slalomzacharyd thank you for the patch!

PR #721, opened 2015-05-27
fix for #718: used service_name instead package_name

igalic:
fix a mistake introduced in #718.


Spredzy:
Duplicate #719

igalic:
thanks, closing this, since travis hasn't given us a green :vertical_traffic_light:  ;)

PR #722, opened 2015-05-28
length check for usernames should take mysql version into consideration

igalic:
Starting MariaDB 10.0.0, usernames are now 80 long.
Our mysql_user and mysql_grant types now take that into consideration.

### DIFF
--- a/lib/puppet/type/mysql_grant.rb
+++ b/lib/puppet/type/mysql_grant.rb
@@ -78,7 +78,14 @@ Puppet::Type.newtype(:mysql_grant) do
         raise(ArgumentError, "Invalid database user #{value}")
       end
 
-      raise(ArgumentError, 'MySQL usernames are limited to a maximum of 16 characters') unless user_part.size <= 16
+      mysql_version = Facter.value(:mysql_version)
+      unless mysql_version.nil?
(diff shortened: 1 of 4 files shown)


igalic:
this still needs tests, of course.
i briefly looked at our types and provider tests, and they are… *sigh* slightly inconsistent, to say the least. Also, right now, we're not even testing for the length, so…

lib/puppet/type/mysql_grant.rb
```diff
@@ -78,7 +78,11 @@ def initialize(*args)
         raise(ArgumentError, "Invalid database user #{value}")
       end
 
-      raise(ArgumentError, 'MySQL usernames are limited to a maximum of 16 characters') unless user_part.size <= 16
+      if Puppet::Util::Package.versioncmp(Facter.value(:mysql_version), '10.0.0') < 0 and user_part.size >= 16
```
hunner: If mysql is not installed and mysql_version is unset, what happens?

igalic: then `versioncmp`'s first argument will be nil, and it will return
```
NoMethodError: undefined method `scan' for nil:NilClass
```

igalic:
FAQ: Why aren't we trying harder to determine the version on/before the first run?
Answer: Why aren't we checking the length of the database?

mhaskel:
:+1: thanks @igalic

PR #723, opened 2015-06-08
Requirement on server package should use 'mysql-server'

xbezdick:
The resouce providing package is called mysql-server, depending on
package name does not seem to work with puppet 4.1.0.


sbadia:
:+1: 
The package name is configured inside the package resource https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/server/install.pp#L9 we must use a static name for deps.
Thanks Lukas!

paramite:
+1

sbadia:
This patch was submitted 12days ago, but it seems that the current change was merged in #727 therefore, we can close this PR, thanks anyway @xbezdick !

PR #727, opened 2015-06-18
(maint) Package was renamed in server::install  - As a result of rename several underlying dependencies broke   this is to update the require dependencies

cyberious:
Change was located here, https://github.com/puppetlabs/puppetlabs-mysql/commit/f5a693b82628983b80a33c0dfc25d633c41001c8

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -19,7 +19,7 @@ class mysql::server::installdb {
       creates   => "${datadir}/mysql",
       logoutput => on_failure,
       path      => '/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin',
-      require   => Package[$mysql::server::package_name],
+      require   => Package['mysql-server'],
     }
 
     if $mysql::server::restart {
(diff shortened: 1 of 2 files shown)


manifests/server/installdb.pp
```diff
@@ -19,7 +19,7 @@
       creates   => "${datadir}/mysql",
       logoutput => on_failure,
       path      => '/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin',
-      require   => Package[$mysql::server::package_name],
+      require   => Package['mysql-server'],
```
hunner: This will still fail if `$mysql::server::package_manage` is false, since the package resource won't be in the catalog.

bmjen: But if `$mysql::server::package_manage` is false or undef and the package resource isn't created, then this should rightfully fail right? But looking at the code, it is set from `$mysql::params::server_package_name` which should always be set.

DavidS: installdb itself is also wrapped in a `if $mysql::server::package_manage`. While this does make that var's name questionable, it does keep it from failing.

The service does not have that protection, and should have a similar construct as the `if $mysql::server::manage_config_file` at https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/server/service.pp#L39

cyberious: As it should fail if the server package is not installed. Perhaps a rework of required with a 
```
notify{'mysql-server': }
...
package{$mysql::package_name:
before => Notify['mysql-server']
}
```
this way others can subscribe to the class and have ordering

PR #729, opened 2015-06-25
Add Solaris support to MySQL module

drewfisher314:
This adds support for Solaris 11.2, 11.3 and 12.0 to the puppetlabs-mysql module.


nibalizer:
Thanks for the contribution!

PR #730, opened 2015-06-25
Don't explode if macaddress isn't set

binford2k:
Sometimes the macaddress fact fails, for example sometimes in a Docker
container. We shouldn't clutter up reports with spurious warnings in
that case.


igalic:
thanks @binford2k!

PR #732, opened 2015-07-14
(maint) don't print diff root password in diff

DavidS:
For puppet versions that support it, suppress diffing root's password.

### DIFF
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -17,6 +17,10 @@ class mysql::server::root_password {
       owner   => 'root',
       mode    => '0600',
     }
+    # show_diff was added with puppet 3.0
+    if versioncmp($::puppetversion, '3.0') <= 0 {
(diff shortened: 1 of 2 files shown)


manifests/server/root_password.pp
```diff
@@ -17,6 +17,10 @@
       owner   => 'root',
       mode    => '0600',
     }
+    # show_diff was added with puppet 3.0
+    if versioncmp($::puppetversion, '3.0') <= 0 {
+      File["${::root_home}/.my.cnf"] { show_diff => false }
```
igalic: nice. i didn't realize you could add defaults for *specific* resources.

DavidS: this is not a default. this is actually setting the value and will cause a conflict if done twice.

igalic: oh! thanks for the clarification

PR #736, opened 2015-07-30
[#puppethack] do not require mysql::server when using mysql::db

igalic:
this is a rebase and fix up of https://github.com/puppetlabs/puppetlabs-mysql/pull/586
this pr closes MODULES-1685.


danieldreier:
+1, this would be extremely useful to me. The new RDS functionality in puppetlabs-aws is effectively useless without this.

danielparks:
+1 I like it too!

PR #737, opened 2015-08-03
Add datefile option to specify whether date should be append to filename

tremblaysimon:
Sometimes you want to be able to do an incremental backup (external app) so your filename must not be changed between each backup if you don't want to break backup incrementality.

I added an option to be able to do that.


igalic:
i feel like i'm not smart enough to understand this change.
for one, for the other, i think you'll have to add the option to all mysql::backup classes.

and finally, this is missing docs and tests

tremblaysimon:
Thanks for your comments, I will add docs and tests. 

I think it's not all mysql::backup classes that have same parameters (I don't know if it is intentional by the way). Since it's a parameter that concern only the template of mysqldump bash script it's why I only added to mysqldump "provider".

bmjen:
ping @tremblaysimon. Thanks for the contribution. Can you please reopen once the requested updates, tests, and docs are done? Thanks!

PR #738, opened 2015-08-03
(MODULES-2051) Suppress diff output for root password

cyberious:
- ensure that we are not logging output of config file
  that contains sensitive password info


igalic:
i think this was already merged in #743

cyberious:
This is a change that must be done, if it is being managed outside oft he mysql scope for another user the change needs to happen in the upstream not in the core module.  I would also be fine with allowing the user to opt out of this instead

PR #741, opened 2015-08-04
Ensure idempotency between Puppet runs

EmilienM:
Running RHEL 7.1, puppetlabs-mysql is not idempotent.
After a second run, puppet tries to change the SElinux context:
/File[mysql-config-file]/seltype: seltype changed 'etc_t' to
'mysqld_etc_t'

Since packaging already manages SElinux labels, we should not let
Puppet doing it (default behavior).
This patch aims to set selinux_ignore_defaults to True for the
mysql-config-file File resource.

Thanks to that patch, Puppet will be indempotent between all runs on
RHEL platforms.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -37,9 +37,10 @@ class mysql::server::config {
 
   if $mysql::server::manage_config_file  {
     file { 'mysql-config-file':
-      path    => $mysql::server::config_file,
-      content => template('mysql/my.cnf.erb'),
-      mode    => '0644',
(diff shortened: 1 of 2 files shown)


hunner:
Could you please submit this from a branch on your own fork instead of from puppetlabs/* ? Thanks.

EmilienM:
@hunner sure, sorry for that.

EmilienM:
done: https://github.com/puppetlabs/puppetlabs-mysql/pull/742

PR #745, opened 2015-08-06
#2030 Only establish dependency between service and package if package is managed.

jonnytpuppet:


danzilio:
Isn't the wording a little awkward? Most places this parameter is called `manage_package` not `package_manage`

jonnytpuppet:
This PR doesn't add or modify the param, just makes use of it properly.

danzilio:
I was just about to update my comment with that ;)

PR #749, opened 2015-08-12
Fixes improper use of function 'warn' in backup manifest of server.

Herr-Herner:
Have a look to the following sites:
https://ask.puppetlabs.com/question/13288/whats-the-difference-between-warn-and-warning/
https://docs.puppetlabs.com/puppet/latest/reference/lang_conditional.html?_ga=1.41504802.1433359326.1439356343


DavidS:
Thanks!

PR #751, opened 2015-08-19
Fix when not managing config file

mcanevet:
This code have been remove here https://github.com/puppetlabs/puppetlabs-mysql/commit/4bab65edcb98f82f87a4414840fe90ab81b6cea3#diff-0938042fe2382aeb10032aa7f8444995 but it is required otherwise we get an `Evaluation Error: Unknown variable: 'install_db_args'.` when using `manage_config_file = false`

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -11,7 +11,8 @@ class mysql::server::installdb {
 
     if $mysql::server::manage_config_file {
       $install_db_args = "--basedir=${basedir} --defaults-extra-file=${config_file} --datadir=${datadir} --user=${mysqluser}"
-
+    } else {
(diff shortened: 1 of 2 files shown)


DavidS:
There is no test with `manage_config_file=false`. That's why this was not noticed before. Hint, hint, nudge, nudge!

igalic:
@mcanevet hint hint nudge nudge

mcanevet:
@DavidS @igalic unit test added

DavidS:
That unit test will succeed even without the change.

mcanevet:
@davids it should fail when using STRICT_VARIABLES="yes", no?

DavidS:
Ah yes, you are, of course correct. Can you squash this into a single commit, so we can merge this?

mcanevet:
@davids done

igalic:
thanks everyone \o/

PR #752, opened 2015-08-24
Have config changes restart mysql.

cdenneen:
For example change bind-address and service still required manual restart.


DavidS:
Please just use the `restart` parameter on the `mysql::server` class.

https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/server.pp#L72

https://github.com/puppetlabs/puppetlabs-mysql#restart

PR #760, opened 2015-09-30
MODULES-2650 Add support for renamed password column

roman-mueller:


igalic:
@roman-mueller, i realize there's an associated ticket, but would you mind elaborating a bit more on the motivation and the nature of the fix in your commit message?

roman-mueller:
@igalic sure.
With MySQL 5.7.6 the "password" column has been renamed to "authentication_string".
See: https://dev.mysql.com/doc/refman/5.7/en/upgrading-from-previous-series.html

Due to this, the query to fetch the password and verify that it is the correct one for this user fails.
This pull request fixes this in a backwards compatible manner:
It will `SELECT authentication_string AS password` only if the version is 5.7.6 or greater.
If a smaller version is used it will `SELECT password` like before.

igalic:
@roman-mueller i actually meant git commit --amend the commit message ;)

roman-mueller:
@igalic I must have missed that part, sorry about that :)

igalic:
thank you @roman-mueller!

PR #762, opened 2015-10-06
changes required to enable replication

markasammut:


manifests/server/replication.pp
```diff
@@ -0,0 +1,41 @@
+#
+class mysql::server::replication {
+  $master_slave  = $mysql::server::master_slave
+  $server_id     = $mysql::server::rep_server_id
+
+  if $mysql::server::replication_enable {
```
DavidS: Please make the inclusion of the replication class optional in the `mysql::server`, instead of protecting the contents. This makes it easier to compose manually, in those ugly edge cases.

manifests/server/replication.pp
```diff
@@ -0,0 +1,41 @@
...
+      ensure  => present,
+      path    => '/usr/my.cnf',
+      section => 'mysqld',
+      setting => "rpl_semi_sync_${master_slave}_enabled",
+      value   => '1',
+      require => Exec['install-replication-plugin'],
+      notify  => Exec['restart-service'],
+    }
+    
+    exec { 'restart-service':
```
DavidS: Please match this to the pattern of https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/server/installdb.pp#L26-L30 to make this work together with the rest of the module across all platforms.

manifests/server/replication.pp
```diff
@@ -0,0 +1,41 @@
+#
+class mysql::server::replication {
+  $master_slave  = $mysql::server::master_slave
+  $server_id     = $mysql::server::rep_server_id
+
+  if $mysql::server::replication_enable {
+    
+    exec { 'install-replication-plugin':
+      command => "mysql -u root -e \"INSTALL PLUGIN rpl_semi_sync_${mysql::server::master_slave} SONAME 'semisync_${mysql::server::master_slave}.so';\"",
```
DavidS: https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/params.pp has a variable for where the `my.cnf` file is located. Please use that instead of hardcoding a single location.

manifests/server/installdb.pp
```diff
@@ -14,12 +14,21 @@
 
     }
 
+    exec { 'remove-replication-params':
+      command => "sed -i '/server-id/ d' /usr/my.cnf; sed -i '/rpl_semi_sync_/ d' /usr/my.cnf",
```
DavidS: https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/params.pp has a variable for where the my.cnf file is located. Please use that instead of hardcoding a single location.

DavidS:
Hi @markasammut ,

Thanks for your contribution! I've added a few comments to the code. Generally this would be a good change.

Please make sure that the additional functionality is also documented in the README and that there are at least basic tests in place.

Also, for final inclusion, please squash all you commits into one, so that the git history remains readable:  https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

If you have experience with replication, it'd be great if you could have a look at #765 and chime in there.

Regards, David

bmjen:
Hi @markasammut,

I'm going to close this PR as stale, can you reopen when you are available to address the issues pointed out by @DavidS above? Thanks!

PR #763, opened 2015-10-07
(MODULES-2676) Support MySQL 5.7.6+

elconas:
Fixes Password Handling for MySQL 5.7.6++
Fixes mysql_install_db for MySQL 5.7.6++
Fixes removed myisam-recove on MySQL 5.7.6++
Added Support for mysqld_version 
Added Provider to create data_directory (with version detection)

MySQL 5.7.6 changed password related SQL commands which needed fixing (done in a prior pull request already), however introducing explicit versioning seems required to support MySQL 5.7 as there are many more incompatible changes (e.g. init db, see below). Also the provider can be used to have better "rspec" tests.

Also MySQL 5.7.6 introduces mysql --initialize-insecure instead of mysql_install_db. To support this without "exec | unless mysqld -V | grep 5.7...." a provider has been written. Also the mysqld_version attribute was added to the base provider to support general version dependent code. 

MySQL 5.7.6 also removed myisam-recover, thus this parameter needs to go to version specific [mysqld-5.X] parameter sections. 

This change has been tested on CentOS 6.7 with MySQL Community Editions: 

  - MySQL Community 5.5 (5.5.46)
  - MySQL Community 5.6 (5.6.27)
  - MySQL Community 5.7 (5.7.8)

Regards, 
Robert

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ replicate-do-db = base1
 replicate-do-db = base2
 ~~~
 
+To implement version specific parameters you can use [mysqld-5.5] syntax which is only read by MySQL version 5.5. This allows one config for different versions of MySQL.
+
 ### Creating a database
 
 To use `mysql::db` to create a database with a user and assign some privileges:
@@ -168,6 +170,7 @@ When working with a remote server, do *not* use the
 #### Private classes
 
 * `mysql::server::install`: Installs packages.
+* `mysql::server::installdb`: Implements setup of mysqld data directory (e.g. /var/lib/mysql)
 * `mysql::server::config`: Configures MYSQL.
 * `mysql::server::service`: Manages service.
 * `mysql::server::account_security`: Deletes default MySQL accounts.
@@ -791,6 +794,17 @@ The name of the MySQL plugin to manage.
 #####  `soname`
 
 The library file name.
+
+#### `mysql_datadir`
+
+Initializes the MySQL data directory with version specific code. Pre MySQL 5.7.6
+it uses mysql_install_db. After MySQL 5.7.6 it uses mysqld --initialize-insecure.
+
+Insecure initialization is needed, as mysqld version 5.7 introduced "secure by default" mode.
+This means MySQL generates a random password and writes it to STDOUT. This means puppet
+can never accesss the database server afterwards, as no credencials are available.
+
+This type is an internal type and should not be called directly.
 
 ### Facts
 
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -3,6 +3,7 @@ class Puppet::Provider::Mysql < Puppet::Provider
   # Without initvars commands won't work.
   initvars
   commands :mysql      => 'mysql'
+  commands :mysqld     => 'mysqld'
   commands :mysqladmin => 'mysqladmin'
 
   # Optional defaults file
@@ -13,7 +14,41 @@ class Puppet::Provider::Mysql < Puppet::Provider
       nil
     end
   end
-  
+
+  def self.mysqld_type
+    # find the mysql "dialect" like mariadb / mysql etc.
+    mysqld_version_string.scan(/\s\(mariadb/i) { return "mariadb" }
+    mysqld_version_string.scan(/\s\(mysql/i) { return "mysql" }
+    mysqld_version_string.scan(/\s\(percona/i) { return "percona" }
+    nil
+  end
+
+  def mysqld_type
+    self.class.mysqld_type
+  end
+
+  def self.mysqld_version_string
+    # we cache the result ...
+    return @mysqld_version_string unless @mysqld_version_string.nil?
+    @mysqld_version_string = mysqld(['-V'].compact)
+    return @mysqld_version_string
+  end
+
+  def mysqld_version_string
+    self.class.mysqld_version_string
+  end
+
+  def self.mysqld_version
+    # note: be prepared for '5.7.6-rc-log' etc results
+    #       versioncmp detects 5.7.6-log to be newer then 5.7.6
+    #       this is why we need the trimming.
+    mysqld_version_string.scan(/\d+\.\d+\.\d+/).first unless mysqld_version_string.nil?
+  end
+
+  def mysqld_version
+    self.class.mysqld_version
+  end
+
   def defaults_file
     self.class.defaults_file
   end
--- a/lib/puppet/provider/mysql_datadir/mysql.rb
+++ b/lib/puppet/provider/mysql_datadir/mysql.rb
@@ -1,0 +1,65 @@
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'mysql'))
+Puppet::Type.type(:mysql_datadir).provide(:mysql, :parent => Puppet::Provider::Mysql) do
+
+  desc 'manage data directories for mysql instances'
+
+  commands :mysqld => 'mysqld'
+  commands :mysql_install_db => 'mysql_install_db'
+
+  def create
+    name                     = @resource[:name]
+    insecure                 = @resource.value(:insecure) || true
+    defaults_extra_file      = @resource.value(:defaults_extra_file)
+    user                     = @resource.value(:user) || "mysql"
+    basedir                  = @resource.value(:basedir) || "/usr"
+    datadir                  = @resource.value(:datadir) || @resource[:name]
+
+    unless defaults_extra_file.nil?
+      if File.exist?(defaults_extra_file)
+        defaults_extra_file="--defaults-extra-file=#{defaults_extra_file}"
+      else
+        raise ArgumentError, "Defaults-extra-file #{defaults_extra_file} is missing"
+      end
+    end
+
+    if insecure == true
+      initialize="--initialize-insecure"
+    else
+      initialize="--initialize"
+    end
+
+    if mysqld_version.nil?
+      debug("Installing MySQL data directory with mysql_install_db --basedir=#{basedir} #{defaults_extra_file} --datadir=#{datadir} --user=#{user}")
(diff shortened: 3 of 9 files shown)


lib/puppet/provider/mysql.rb
```diff
@@ -13,7 +14,22 @@ def self.defaults_file
...
     end
   end
-  
+
+  def self.mysqld_version
+    # we cache the result ...
+    # note: be prepared for '5.7.6-rc-log' etc results
+    #       versioncmp detects 5.7.6-log to be newer then 5.7.6
+    #       this is why we need the trimming.
+    return @mysqld_version_string.scan(/\d+\.\d+\.\d+/).first unless @mysqld_version_string.nil?
```
DavidS: Shouldn't this regex (and the one below) be anchored with `^`, or at least have an additional space in front to not accidentally match something else?

lib/puppet/type/mysql_datadir.rb
```diff
@@ -0,0 +1,31 @@
+Puppet::Type.newtype(:mysql_datadir) do
+  @doc = 'Manage MySQL datadirs with mysql_install_db OR msqld (5.7.6 and above).'
```
DavidS: s/msqld/mysqld/

lib/puppet/type/mysql_datadir.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  newparam(:basedir) do
+    desc 'The basedir name, default /usr.'
+    newvalues(/^\//)
+  end
+
+  newparam(:user) do
+    desc 'The user for the directory default mysql (name, not uid).'
+    newvalues(/\D+/)
```
DavidS: user names can contain numbers

spec/unit/puppet/provider/mysql_user/mysql_spec.rb
```diff
@@ -85,6 +98,38 @@
...
+       expect(Puppet::Util::Package.versioncmp(provider.mysqld_version, '5.7.6')).to be == 0
+    end
+    it 'detects >= 5.7.6 for 5.7.7-log' do
+       provider.class.instance_variable_set(:@mysqld_version_string, 'mysqld  Ver 5.7.7-log for Linux on x86_64 (MySQL Community Server (GPL))')
+       expect(Puppet::Util::Package.versioncmp(provider.mysqld_version, '5.7.6')).to be >= 0
+    end
+    it 'detects < 5.7.6 for 5.7.5-log' do
+       provider.class.instance_variable_set(:@mysqld_version_string, 'mysqld  Ver 5.7.5-log for Linux on x86_64 (MySQL Community Server (GPL))')
+       expect(Puppet::Util::Package.versioncmp(provider.mysqld_version, '5.7.6')).to be < 0
+    end
```
DavidS: You could compress this whole block with

```ruby
[
  [ 'mysqld  Ver 5.7.6 for Linux on x86_64 (MySQL Community Server (GPL))', '5.7.6 ]
  # ...
].each do |dash_v, version|
  it "detects '#{dash_v}' as '#{version}'" do
    provider.class.instance_variable_set(:@mysqld_version_string, dash_v)
    expect(provider.mysqld_version).to eq version
  end
end
```

spec/unit/puppet/provider/mysql_user/mysql_spec.rb
```diff
@@ -103,12 +148,19 @@
...
 
       provider.expects(:password_hash).returns('*6C8989366EAF75BB670AD8EA7A7FC1176A95CEF5')
       provider.password_hash=('*6C8989366EAF75BB670AD8EA7A7FC1176A95CEF5')
     end
+    it 'changes the hash >= 5.7.6' do
+      provider.class.instance_variable_set(:@mysqld_version_string, '5.7.6')
+      provider.expects(:mysql).with([defaults_file, '-e', "ALTER USER 'joe'@'localhost' IDENTIFIED WITH mysql_native_password AS '*6C8989366EAF75BB670AD8EA7A7FC1176A95CEF5'"]).returns('0')
+
+      provider.expects(:password_hash).returns('*6C8989366EAF75BB670AD8EA7A7FC1176A95CEF5')
+      provider.password_hash=('*6C8989366EAF75BB670AD8EA7A7FC1176A95CEF5')
```
DavidS: When you're already in there, at the top of the tests, there is a `newhash` defined which has this value, I think this should use that then?

DavidS:
Hi, thank you for your contribution. I've added only a few minor comments to the code. Additionally it requires a few sentences in the README's reference section explaining the new type.

DavidS:
PS: @igalic , please also have a look at this!

elconas:
Hello, fixed according to your feedback. 

Also I fixed MariaDB 10 regression (which is actually 5.6, but is detected as > 5.7.6). Now I added "flavor" detection and made the 5.7.6 changes very closely coupled with MySQL type.

README.md
```diff
@@ -792,6 +795,17 @@ The name of the MySQL plugin to manage.
 
 The library file name.
 
+#### `mysql_datadir`
+
+Initializes the MySQL data directory with version specific code. Pre MySQL 5.7.6
+it uses mysql_install_db. After MySQL 5.7.6 it uses mysqld --initialize-insecure.
+
+Insecure initialization is needed, as mysqld version 5.7 indroduced "secure by default" mode.
```
DavidS: s/indroduced/introduced

lib/puppet/provider/mysql_datadir/mysql.rb
```diff
@@ -0,0 +1,65 @@
...
+Puppet::Type.type(:mysql_datadir).provide(:mysql, :parent => Puppet::Provider::Mysql) do
+
+  desc 'manage data directories for mysql instances'
+
+  commands :mysqld => 'mysqld'
+  commands :mysql_install_db => 'mysql_install_db'
+
+  def create
+    name                     = @resource[:name]
+    insecure                 = @resource.value(:insecure) || :true
```
DavidS: Please use just plain booleans here instead of `:true`/`:false`. Puppet core has even a separate boolean property: https://docs.puppetlabs.com/guides/custom_types.html#boolean-parameters

lib/puppet/provider/mysql_datadir/mysql.rb
```diff
@@ -0,0 +1,65 @@
...
+      if mysqld_type == "mysql" and Puppet::Util::Package.versioncmp(mysqld_version, '5.7.6') >= 0
+        debug("Initializing MySQL data directory >= 5.7.6 with 'mysqld #{defaults_extra_file} #{initialize} --basedir=#{basedir} --datadir=#{datadir} --user=#{user}'")
+        mysqld([defaults_extra_file,initialize,"--basedir=#{basedir}","--datadir=#{datadir}", "--user=#{user}", "--log_error=/var/tmp/mysqld_initialize.log"].compact)
+      else
+        debug("Installing MySQL data directory with mysql_install_db --basedir=#{basedir} #{defaults_extra_file} --datadir=#{datadir} --user=#{user}")
+        mysql_install_db(["--basedir=#{basedir}",defaults_extra_file, "--datadir=#{datadir}", "--user=#{user}"].compact)
+      end
+    end
+
+   exists? ? (return true) : (return false)
```
DavidS: Since `exists?` already returns a bool, you can just say `exists?` instead of funging around with the ternary operator here.

DavidS:
Since this is a larger change, I'd be very thankful if someone else, like @igalic , @roman-mueller or @mcanevet , could have a look.

Meanwhile, it'd be great if you could rebase/squash this into a single commit to avoid cluttering the git history. (See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits ).

DavidS:
PS: Adding `(MODULES-2676)` into the commit's title will automatically link this back to the JIRA ticket.

mcanevet:
This is a scary one... @elconas would it be possible to merge this step by step? For example, a first PR that replaces `Exec[mysql_install_db]` by the `mysql_datadir` resource. Then it would be easier to add MySQL 5.7.6++ support.

elconas:
Hi, this is difficult - from my perspective - as I need Versioning (5.5.X) and Typing (mysql/percona/mariadb) support for replacing the Exec[mysql_install_db]. So if I have Versioning and Typing, adding the two different version specific SQL statements in the user management section of the manifest is kind of trivial. 

This makes the replace Exec containing 90% of the change, making this change not really splittable.

elconas:
Did in same branch, to keep the pull request and the history: 

git rebase -i 2a8693482f962b36024e9c70a459de8e9dc65eb6
git push origin +fix_mysql576_pull

DavidS:
I'm now running it locally through some tests and will merge unless something pops up.

DavidS:
@elconas the tests fail with `Error: Could not find a suitable provider for mysql_datadir` as `mysqld` is not a command on the path on Centos 6.4. Instead it is placed into `/usr/libexec`. Since the agent doesn't look for it there, it marks the provider as invalid and fails.

mentat:
@elconas can you rebase this and ensure the tests pass?

spec/unit/puppet/provider/mysql_user/mysql_spec.rb
igalic: :purple_heart:

igalic:
@mentat the centos7 tests can, for now, be ignored.

mentat:
Hi @elconas could you rebase this one so we can get it merged?

DavidS:
I've rebased this to #786 and addressed the merge conflicts there.

DavidS:
and it was merged. Thanks for all your work!

mentat:
Hi @elconas it seems this PR fails for redhat 6 & 7 but works on Ubuntu.  Do you have any ideas how to improve the compatibility?

elconas:
Hi, Fixed in https://github.com/puppetlabs/puppetlabs-mysql/pull/789

PR #765, opened 2015-10-07
Ticket Forge Modules / MODULES-2675 - Unique server_id

jmk61:
Ticket Forge Modules / MODULES-2675
Fixed Unique server_id facter fact. Allowing a unique id to be generated by reading in all 6 splits of the mac address, where as it was only reading in 4 of the 6 splits.

### DIFF
--- a/lib/facter/mysql_server_id.rb
+++ b/lib/facter/mysql_server_id.rb
@@ -1,5 +1,5 @@
 def get_mysql_id
-  Facter.value(:macaddress).split(':').inject(0) { |total,value| (total << 4) + value.hex }
+  Facter.value(:macaddress).split(':').inject(0) { |total,value| (total << 6) + value.hex }
 end
 
 Facter.add("mysql_server_id") do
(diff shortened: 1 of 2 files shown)


DavidS:
Hi @jmk61 , thanks for your contribution! Would you be son nice and squash this into a single commit, so as to keep the history clean?

See Squashing: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

DavidS:
A more difficult question: how is this fact used, and how will changing it's output impact the users of this?

jmk61:
The fact is used to determine a unique server ID within the my.cnf which allows master-slave setups. By modifying the fact it will change the users server_id within their conf. It shouldn't have a direct effect on existing setups as I'm unware of a way to directly utilize this information within a setup other then distinguishing each mysql server in your setup.

Please let me know your thoughts, questions and concerns.

Thanks

Joseph Karns
System Administrator III
Penn State University
Outreach Information Technology
329 Innovation Blvd
Suite 306
State College, Pa 16803

----- Reply message -----

jmk61:
David:

Not sure if I did the squash correct. If not request the pull request and I'll resubmit.

Thanks
Joey

DavidS:
Yeah, this branch doesn't look like it should. Also, I'm still not 100% clear what happens to an existing replication setup, when all the `server_id`s change. The docs don't seem to address that.

DavidS:
Closing in preference to #767

PR #767, opened 2015-10-12
Fixes unique server_id within my.cnf Ticket/MODULES-2675

jmk61:
Fixes unique server_id within my.cnf, The issue was that the entire mac address was not being read in to generate the id.

David...here is the new pull request. I'm still not sure how we can test this bug fix.


DavidS:
A valid test would be to set up a mysql cluster with the old module, then change the server ids and verify that the replication is still functional.

PR #769, opened 2015-10-15
Ubuntu vivid should use systemd not upstart

gabriel403:


manifests/params.pp
```diff
@@ -351,7 +351,14 @@
 
   case $::operatingsystem {
     'Ubuntu': {
-      $server_service_provider = upstart
```
igalic: server is the only thing that runs a service, so it's kinda redundant.

manifests/params.pp
```diff
@@ -351,7 +351,14 @@
...
+      case $::lsbdistcodename {
+        'vivid': {
+          $server_service_provider = systemd
+        }
+        default: {
+          $server_service_provider = upstart
+        }
+      }
```
igalic: and this whole thing can be replaced with a selector:
```puppet
$server_service_provider = $::lsbdistcodename {
     'vivid' => 'systemd',
     default: 'upstart'
}
```
and now the problem becomes extending it for future versions…

i'm really starting to feel the burden of the systemd breakage in debuntu… :(

gabriel403: @igalic thanks for the tip! Will implement

DavidS:
Please don't use codenames for these checks, but a open range on the release version. We can expect that from now on all Ubuntu releases will use systemd.

gabriel403:
@DavidS would be glad to, but you're gonna have to point me in the right direction for that

gabriel403:
@DavidS `versioncmp($::lsbdistrelease, '14.10') == 1` ?

DavidS:
Almost :-) `lsbdistrelease` is not available always, but `operatingsystemmajrelease` is. to aid readability use `> 0 `:

```
versioncmp($::operatingsystemmajrelease, '14.10') > 0
```

bmjen:
Hi @gabriel403, any progress on this PR?

gabriel403:
@bmjen hi sorry! Will sort this tonight :)

gabriel403:
@bmjen @DavidS sorry about the huge delay on this, should be sorted now, if you can validate it's as you expect I'll rebase and squash

DavidS:
@gabriel403 looks good, except for the missing `{` :

```
Syntax error at 'server_service_provider'; expected '}' at /home/travis/build/puppetlabs/puppetlabs-mysql/manifests/params.pp:363
```

gabriel403:
@DavidS looks like we have a winner? Anything else you'd like me to do?

DavidS:
Celebrate? :-)  Thanks!

gabriel403:
Woo! Thanks @DavidS

PR #770, opened 2015-10-16
Add ability to input password hash directly

emahags:


mentat:
Hey this looks great, would you mind adding the password_is_hash documentation to the https://github.com/emahags/puppetlabs-mysql/blob/2702-password_hash/README.md#mysql_user section?

hunner:
Would it perhaps make more sense to do this automatically by matching `password` against a regex that knows if it is already a hash or not? Then we wouldn't need `password_is_hash`

You wouldn't be able to submit passwords that are literally a hash, but if you're doing that then you're just odd :D

emahags:
@mentat, @hunner: While working on your comments I thought of another way to solve this problem. Since there are other places where this hashing fix could be applied (root password for example) I went to the source function instead and did this modification: https://github.com/emahags/puppetlabs-mysql/commit/ee94b3ead1182c37eec83ff1aaff714a7c63a895

Would this be an ok solution? It changes the scope quite a bit and I'm not really sure where/how to document it.

mentat:
I like that idea, it seems that the regex is strong enough to keep out accidental uses.  If we document this well we should be ok.  @hunner?

pckls:
https://tickets.puppetlabs.com/browse/MODULES-2028

mentat:
Can you update this pull request and squash it down to a single commit?

emahags:
@mentat Done. I wasn't sure where to document this, so I made a general note in the readme, I hope it's ok.

jonnytpuppet:
Hi There,
It would be great if you could have a look at the failing tests.
Thanks!

emahags:
@jonnytpuppet The test was already failing before my commit though. See https://travis-ci.org/emahags/puppetlabs-mysql/builds/87403257

bmjen:
@emahags it appears that the test failures are due to other reasons, however I think those failures have been fixed in master. Do you mind doing a rebase off master, so any unit test fixes can be pulled in and run?

emahags:
@bmjen Rebase done. I found some issues though, if you use a hashed password the module seems to switch between it and a faulty password on every puppet run. I'm trying to find out the reason but no luck so far, any tips?

emahags:
Hmm, I can't seem to reproduce that bug anymore. I saw it before the rebase, so maybe it want away with the new commits.

PR #775, opened 2015-11-06
(MODULES-2605) Use MYSQL_PWD to avoid mysqldump warnings.

abednarik:
Instead of using password as a command arguments, we are setting MYSQL_ENV variable. See https://dev.mysql.com/doc/refman/5.1/en/password-security-user.html


igalic:
thanks, @abednarik!

PR #776, opened 2015-11-07
(MODULES-2767) allow to check if table exists before grant

pondohva:


lib/puppet/parser/functions/mysql_table_exists.rb
```diff
@@ -0,0 +1,39 @@
...
+      if db_name.eql?('*') or table_name.eql?('*')
+        true
+      else
+        query = "SELECT TABLE_NAME FROM information_schema.tables WHERE TABLE_NAME = '#{table_name}' AND TABLE_SCHEMA = '#{db_name}';"
+
+        %x{mysql #{defaults_file} -NBe #{query}}.strip.eql?(table_name)
+      end
+    else
+      raise(Puppet::ParseError, 'mysql_table_exists() accept 1 argument - table string like \'database_name.table_name\'')
+    end
```
igalic: we should revert this whole thing for early returns:

```ruby
return raise(Puppet::ParseError, "mysql_table_exists() accept 1 argument - table string like 'database_name.table_name'") unless match = args[0].match(/(.*)\.(.*)/)
db_name, table_name = match.captures
return true if (db_name.eql?('*') or table_name.eql?('*')) # this is actually not true
query = "SELECT TABLE_NAME FROM information_schema, tables WHERE TABLE_NAME etc.."
%x{mysql etc..}
```

pondohva:
@igalic thank you for pointing this out! Fixed

igalic:
thank you @pondohva!

pondohva:
thank you, @igalic . Now we have problem with rspec test - https://travis-ci.org/puppetlabs/puppetlabs-mysql/builds/90054252 (it's my fault, sorry), and I've forgotten check for args.size in function body.
I'll create PR to fix this ASAP :)

PR #777, opened 2015-11-09
(MODULES-2767) fix mysql_table_exists: add check for args.size, fix rspec test

pondohva:
https://github.com/puppetlabs/puppetlabs-mysql/pull/776

rspec test doesn't fail now.


igalic:
looks good… well, on travis' side, anyway.
pcci seems to be hosed in general.

PR #779, opened 2015-11-15
Fixes edge-case with dropping pre-existing users with grants

jmcclell:
If a user exists in the database upon first Puppet run (for example, in the case of loading a database snapshot) and the run sets that user's :ensure attribute to 'absent', the mysql_grant provider will throw an error when the dependency chain causes it to try to destroy the grants associated with that user because the DROP statement from the mysql_user provider already removed the grants. To fix, we must check if the user exists before revoking the grants.

### DIFF
--- a/lib/puppet/provider/mysql_grant/mysql.rb
+++ b/lib/puppet/provider/mysql_grant/mysql.rb
@@ -114,7 +114,11 @@ Puppet::Type.type(:mysql_grant).provide(:mysql, :parent => Puppet::Provider::Mys
   end
 
   def destroy
-    revoke(@property_hash[:user], @property_hash[:table])
+    # if the user was dropped, it'll have been removed from the user hash
+    # as the grants are alraedy removed by the DROP statement
+    if self.class.users.include? @property_hash[:user]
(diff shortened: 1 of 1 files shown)


jmcclell:
This is potentially inefficient compared to, perhaps, catching the raised error and rescuing it if it matches a particular error pattern, though that feels a bit dirty and it doesn't seem like grants get dropped very often in the vast majority of use cases. Thoughts?

igalic:
@jmcclell is there any way to (acceptance) test for this?

jmcclell:
Without a full integration test (actually having a running mysql with the proper starting condition) I don't know how one could model a test. My ruby-foo isn't very strong, either, I'm afraid.

jonnytpuppet:
@jmcclell This actually looks good to me, and personally, I'd be happy enough if the current grant destroy tests still pass.

It would be appreciated if you could squash your commits though.

Thanks!

jmcclell:
Done.

igalic:
thanks @jmcclell & @jonnytpuppet!

PR #782, opened 2015-11-27
(MODULES-2633) Do not hardcode root user

abednarik:
See comments here https://tickets.puppetlabs.com/browse/MODULES-2633 please.

There are some doubts I have.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The MySQL module manages both the installation and configuration of MySQL, as we
 ### Beginning with MySQL
 
 If you want a server installed with the default options you can run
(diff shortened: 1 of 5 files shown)


igalic:
what about the providers?

jonnytpuppet:
Hi @abednarik 

I like this feature, however it needs to be comprehensive for it to work properly. We would need the providers updated as per @igalic 's comments.

Thanks!

abednarik:
Hi @jonnytpuppet 

 Sorry, but I'm busy with other work at this moment. I will cose this one, since i don't have the time to do that.

Cheers.

PR #787, opened 2015-12-14
Fixing error when disabling service management and the service does not exist

obi11235:
If you set service_manage => false, because your instance of mysql is not managed by any service management OR in my case I'm only using this puppet module to manage mysql users, puppet generates an error.

Error: /Stage[main]/Mysql::Server::Service/Service[mysqld]: Could not evaluate: Could not find init script or upstart conf file for 'mysql'

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -26,11 +26,13 @@ class mysql::server::service {
     }
   }
 
-  service { 'mysqld':
-    ensure   => $service_ensure,
-    name     => $mysql::server::service_name,
-    enable   => $mysql::server::real_service_enabled,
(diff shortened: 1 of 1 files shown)


pabelanger:
+1

Not sure what is wrong with the CI however

igalic:
centos build has been broken since forever
there's a way to fix it, and we should push that into modulesync
ping @nibalizer @blkperl @bkero

roidelapluie:
@igalic please note that this is a breaking change that should trigger more than a minor release version (y.Y.z )

roidelapluie:
(e.g this should be released in 3.7)

PR #788, opened 2015-12-17
Fix CentOS 

elconas:
Fixed CentOS (not nice, but using confines and changing the whole class hierarchy is too complex for me). Using has_command { environment :PATH => .... } does not work too ... so this hack "is it" :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Supported Release 3.6.2
+###Summary
+
+Small release for support of newer PE versions. This increments the version of PE in the metadata.json file.
+
 ## 2015-09-22 - Supported Release 3.6.1
 ### Summary
 This is a security and bugfix release that fixes incorrect username truncation in the munge for the mysql_user type, incorrect function used in `mysql::server::backup` and fixes compatibility issues with PE 3.3.x.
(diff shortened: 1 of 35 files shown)


elconas:
Hi, this failed on CentOS with default MySQL (non extra repos) as /usr/libexec was not part of the path for mysqld.
Fixed in https://github.com/puppetlabs/puppetlabs-mysql/pull/788 and tested on CentOS with default package and upstream mysql-community packages.
Tested on: 

CentOS6:
  - MariaDB 10
  - Oracle MySQL Community Edition 5.5 (mysql-community-server)
  - Oracle MySQL Community Edition 5.6 (mysql-community-server)
  - Oracle MySQL Community Edition 5.7 (mysql-community-server)
  - CentOS default mysql 5.1 (mysql-server)

Also Updated https://tickets.puppetlabs.com/browse/MODULES-2676

igalic:
@elconas could you do this as a proper [rebase](https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing) here?
see also https://github.com/puppet-community/puppet-jira/pull/99#issuecomment-152528108 for how to do that

elconas:
Created new Pull Request https://github.com/puppetlabs/puppetlabs-mysql/pull/789

PR #791, opened 2016-01-08
Re-Add the ability to set a empty string as option parameter

roidelapluie:
Prior to 136b1aa646efde57b91fd994bfcf55b38ba7c294 it was possible to
have an empty string as value of a my.cnf parameter, resulting in a line
with just the parameter name.

That commit re-enable that behavior that was removed by accident.

### DIFF
--- a/spec/classes/mycnf_template_spec.rb
+++ b/spec/classes/mycnf_template_spec.rb
@@ -29,6 +29,11 @@ describe 'mysql::server' do
           end
         end
 
+        describe 'skip-name-resolve set to an empty string' do
+          let(:params) {{ :override_options => { 'mysqld' => { 'skip-name-resolve' => '' }}}}
+          it { is_expected.to contain_file('mysql-config-file').with_content(/^skip-name-resolve$/) }
+        end
+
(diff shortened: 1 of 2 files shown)


igalic:
duh… i think it was discussed separately, but never really fixed :(
thank you very much for catching and fixing this, @roidelapluie!

hunner:
@baurmatt What about using `~` as the value in hiera, assuming you're using yaml? Or whatever the nil/null value is for the backend you're using. re https://groups.google.com/forum/#!topic/puppet-users/0l3m4VCHjOg

baurmatt:
@hunner Woah, honestly i thought no one would care about a comment in an old issue, but you did and suggested a solution which doesn't need any modification at all. Thank you very much, works great! :)

hunner:
I get emailed for comments and usually keep an eye out for PRs that broke something unexpected. Thanks for reporting it!

roidelapluie:
Thanks guys. Sorry it broke your setup, be aware that it broke mine first :) Nice @hunner the ~ tip.

PR #796, opened 2016-02-03
(#3028) Fix mysql_grant with MySQL ANSI_QUOTES mode

jhriggs:
Change mysql_grant provider to ignore/delete double-quotes -- as it does with single quotes and backticks -- in the returned list of existing grants. With ANSI_QUOTES enabled in MySQL's sql_mode, grant identifiers (e.g. database name) are quoted with double-quotes rather than backticks, for example "foo".* vs. `foo`.*. This breaks mysql_grant's evaluation of existing grants and causes it to apply grants with every run.


bmjen:
Thanks for the contribution @jhriggs !

PR #797, opened 2016-02-07
mark Ubuntu xenial as supported

mmoll:
Current master runs fine on xenial (mainly thanks to 681d4f8), a new released version would be nice.


HelenCampbell:
@mmoll Thank you! I've created a ticket to push a release of mysql which will include this.

PR #801, opened 2016-02-18
Generate .my.cnf for all sections

holser:
MySQL ecosystem has a lot of tools that use .my.cnf such as mysqldump or
mysqladmin. Generating all sections allows operator to use them under
root account without passing -p<password>

### DIFF
--- a/templates/my.cnf.pass.erb
+++ b/templates/my.cnf.pass.erb
@@ -1,9 +1,11 @@
 ### MANAGED BY PUPPET ###
 
-[client]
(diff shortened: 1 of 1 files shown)


degorenko:
+1 to this commit

solomonty:
Why is this change necessary? All of the tools use the `[client]` section without needing to explicitly specify the credentials.

[Using Option Files](http://dev.mysql.com/doc/refman/5.6/en/option-files.html) 

> The [client] option group is read by all client programs (but not by mysqld). This enables you to specify options that apply to all clients. For example, [client] is the perfect group to use to specify the password that you use to connect to the server.

PR #803, opened 2016-03-03
remove option from config if override_options set false or 'off'

mklette:
In it's current state the my.cnf-template is not capable of removing a specific boolean option, i.e. 'skip-name-resolve'. Activation of such an option works properly and generates a single line with only the option name listed. If I do pass 'false' instead, the following line would be generated:
    skip-name-resolve = false
which MySQL (5.5) is happily using to activate the option:
```sql
show VARIABLES where Variable_Name = 'skip_name_resolve';
+-------------------+---------+
| Variable_name     | Value   |
|-------------------+---------|
| skip_name_resolve | ON      |
+-------------------+---------+
```
I'd like to solve this by adding a new condition to my.cnf-template to just ignore/skip any option whose content is false or 'off'.
Is this acceptable?

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -44,6 +44,15 @@ class mysql::server::config {
     }
   }
 
+  # Issue a warning if other options are used for disabling SSL: according to
+  # the template they will be ordered before other 'ssl-*' options and in turns
+  # leads MySQL to ignore the deactivation request.
+  if (!$options['mysqld']['ssl'] or $options['mysqld']['skip-ssl'] or $options['mysqld']['disable-ssl']) {
+    notify {'warn-ineffective-disable-options':
+      message => 'Use "ssl-disable" => true to disable SSL! Other options are not respected!'
+    }
+  }
+
   if $options['mysqld']['ssl-disable'] {
     notify {'ssl-disable':
       message =>'Disabling SSL is evil! You should never ever do this except if you are forced to use a mysql version compiled without SSL support'
--- a/spec/classes/mycnf_template_spec.rb
+++ b/spec/classes/mycnf_template_spec.rb
@@ -42,19 +42,25 @@ describe 'mysql::server' do
 
         describe 'ssl set to false' do
           let(:params) {{ :override_options => { 'mysqld' => { 'ssl' => false }}}}
-          it { is_expected.to contain_file('mysql-config-file').with_content(/ssl = false/) }
+          it { is_expected.to contain_notify('warn-ineffective-disable-options') }
         end
 
         # ssl-disable (and ssl) are special cased within mysql.
         describe 'possibility of disabling ssl completely' do
           let(:params) {{ :override_options => { 'mysqld' => { 'ssl' => true, 'ssl-disable' => true }}}}
-          it { is_expected.to contain_file('mysql-config-file').without_content(/ssl = true/) }
+          it { is_expected.to contain_file('mysql-config-file').without_content(/^ssl.*/) }
         end
 
         describe 'a non ssl option set to true' do
           let(:params) {{ :override_options => { 'mysqld' => { 'test' => true }}}}
           it { is_expected.to contain_file('mysql-config-file').with_content(/^test$/) }
           it { is_expected.to contain_file('mysql-config-file').without_content(/test = true/) }
+        end
+
+        describe 'a non ssl option set to false' do
+          let(:params) {{ :override_options => { 'mysqld' => { 'test' => false }}}}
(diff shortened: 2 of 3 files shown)


hunner:
Hah, whoops. Yeah, looks great. Could you make sure there is docs for that parameter? And the unit test at https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/classes/mycnf_template_spec.rb#L32-L35 could be copied and checked with `'off'` to verify.

templates/my.cnf.erb
```diff
@@ -6,8 +6,10 @@
 <%     v.sort.map do |ki, vi| -%>
 <%       if ki == 'ssl-disable' or (ki =~ /^ssl/ and v['ssl-disable'] == true) -%>
 <%         next %>
-<%       elsif vi == true or vi == '' -%>
+<%       elsif [true, '', 'on'].include?(vi) -%>
 <%=        ki %>
+<%       elsif [false, 'off'].include?(vi) -%>
```
hunner: I wonder if we would need `:false` here... sometimes puppet passes symbols instead of booleans.

mklette: Thanks a lot for the first review!!
I guess I'll need to take some more time for this. My change introduces test failures with 'ssl = false'. I did a quick test on this against Ubuntu 14.04 with MySQL 5.5 and couldn't immediatly prove it's of any real use but I really need to test this further. If you don't mind, I'd like to keep this open until I could check this properly.

mklette:
May I ask you for another review?

spec/classes/mycnf_template_spec.rb
```diff
@@ -36,19 +36,24 @@
...
           let(:params) {{ :override_options => { 'mysqld' => { 'ssl' => true }}}}
-          it { is_expected.to contain_file('mysql-config-file').with_content(/ssl/) }
+          it { is_expected.to contain_file('mysql-config-file').with_content(/^ssl$/) }
           it { is_expected.to contain_file('mysql-config-file').without_content(/ssl = true/) }
         end
 
         describe 'ssl set to false' do
           let(:params) {{ :override_options => { 'mysqld' => { 'ssl' => false }}}}
-          it { is_expected.to contain_file('mysql-config-file').with_content(/ssl = false/) }
+          it { is_expected.to contain_file('mysql-config-file').without_content(/^ssl$/) }
```
igalic: the `^anchoring$` here makes me suspicious of false negatives? postitives?
wrong non-matches anyway…

solomonty:
The my.cnf code generation used to work that way.

For example, I have these set to false and they appear as comments in the my.cnf.

```
#log-queries-not-using-indexes
#read_only
#skip-networking
```

manifests/server/config.pp
```diff
@@ -44,6 +44,15 @@
     }
   }
 
+  # Issue a warning if other options are used for disabling SSL: according to
+  # the template they will be ordered before other 'ssl-*' options and in turns
+  # leads MySQL to ignore the deactivation request.
+  if (!$options['mysqld']['ssl'] or $options['mysqld']['skip-ssl'] or $options['mysqld']['disable-ssl']) {
+    notify {'warn-ineffective-disable-options':
+      message => 'Use "ssl-disable" => true to disable SSL! Other options are not respected!'
```
solomonty: This is not necessary. The synonyms work according to the documentation.

https://dev.mysql.com/doc/refman/5.5/en/using-secure-connections.html
https://dev.mysql.com/doc/refman/5.6/en/using-secure-connections.html

> To prevent use of encryption and override other --ssl-xxx options, invoke the client program with --ssl=0 or a synonym (--skip-ssl, --disable-ssl)

https://dev.mysql.com/doc/refman/5.7/en/using-secure-connections.html

> To prevent use of encryption and override other --ssl-xxx options, invoke the client program with --ssl-mode=DISABLED, --ssl=0, or a synonym (--skip-ssl, --disable-ssl)

> To use an unencrypted connection, invoke the client with --ssl-mode=DISABLED, --ssl=0, or a synonym (--skip-ssl, --disable-ssl)

solomonty:
I prefer the former behavior where the option is added to the cnf file as a comment instead of removing it.

pkatiushyn:
I am removing unnecessary options by:
'skip-external-locking'                    => undef,
It works well.

mklette:
I'm closing this pull request for several reasons. Most importantly because of the fact that I don't use this module anymore. Further, I didn't found the time to verify above mentioned comments and probably also wont find it in future. Thanks anyways for all your reviews!

PR #808, opened 2016-03-11
Use mysql_install_db only with uniq defaults-extra-file

mmalchuk:
Executing mysql_install_db with the defaults-extra-file set to the
global option file can produce errors with duplicates. For example it
happens when 'ignore-db-dir' option passed to $override_options in
mysql::server class. Same as https://bugs.mysql.com/bug.php?id=69441

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -9,7 +9,7 @@ class mysql::server::installdb {
     $basedir = $mysql::server::options['mysqld']['basedir']
     $config_file = $mysql::server::config_file
 
-    if $mysql::server::manage_config_file {
+    if $mysql::server::manage_config_file and $config_file != $mysql::params::config_file {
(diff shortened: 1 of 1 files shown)


mmalchuk:
Is it possible to tag this commit as 3.6.3? Or there is a some workflow for this?

hunner:
3.6.2 is the latest release and a 3.7.0 release has already passed review and just waiting for tagging. Our releases are tagged and gpg signed after a review process involving QA, docs, eng, and release eng so it's not as easy as just tagging a commit, unfortunately.

Anyway, the next release should include this fix as the master branch is where release branches are made.

hunner:
In the mean time if you point your r10k config (or however you deploy your modules) at this hash, that would let you use the fix before it is shipped.

mmalchuk:
Thanks a lot.

PR #809, opened 2016-03-13
Use mysql_install_db only with uniq defaults-extra-file

mmalchuk:
Executing mysql_install_db with the defaults-extra-file set to the
global option file can produce errors with duplicates. For example it
happens when 'ignore-db-dir' option passed to $override_options in
mysql::server class. Same as https://bugs.mysql.com/bug.php?id=69441


mmalchuk:
This commit is the same as #808 but for 3.6.x until we switch to the master.

PR #811, opened 2016-03-17
Fixed global parameters skipped

pashamesh:
Fixed problem with ignoring parameters from global configs

### DIFF
--- a/templates/mysqlbackup.sh.erb
+++ b/templates/mysqlbackup.sh.erb
@@ -70,18 +70,18 @@ cleanup
 <% if @file_per_database -%>
 mysql --defaults-file=$TMPFILE -s -r -N -e 'SHOW DATABASES' | while read dbname
 do
-  mysqldump --defaults-file=$TMPFILE --opt --flush-logs --single-transaction \
+  mysqldump --defaults-extra-file=$TMPFILE --opt --flush-logs --single-transaction \
     ${ADDITIONAL_OPTIONS} \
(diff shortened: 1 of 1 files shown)


templates/mysqlbackup.sh.erb
```diff
@@ -70,18 +70,18 @@ cleanup
 <% if @file_per_database -%>
 mysql --defaults-file=$TMPFILE -s -r -N -e 'SHOW DATABASES' | while read dbname
 do
-  mysqldump --defaults-file=$TMPFILE --opt --flush-logs --single-transaction \
+  mysqldump --defaults-extra-file=$TMPFILE --opt --flush-logs --single-transaction \
```
cyberious: What versions of mysql support the `--defaults-extra-file`?

More so what version was it introduced? And what is the use case? Could you elaborate?

pashamesh: `--defaults-extra-file` present in docs since `5.0` http://dev.mysql.com/doc/refman/5.0/en/mysqldump.html#option_mysqldump_defaults-extra-file. I can't find confirmation for earlier versions quickly.

I faced the problem that mysql backup script can't find socket file. It was caused by mysql built from sources.

bmjen: I don't have any issues with this and I'm not too worried about the installations of version < 5.0.

Any objections @cyberious ?

igalic: :+1: from me, anyway

PR #813, opened 2016-03-24
Config before install

tomkrouper:
When there is a new install, the package installation will run `mysqld
--initialize` or `mysql_install_db` which leaves `mysqld` in a running
state. If you have `$mysql::server::restart` set to false, `mysqld` will
not restart and you will be left in a state of running a default config
instead of your intended configuration.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -53,8 +53,8 @@ class mysql::server (
 
   Class['mysql::server::root_password'] -> Mysql::Db <| |>
 
-  include '::mysql::server::install'
   include '::mysql::server::config'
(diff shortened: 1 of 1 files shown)


igalic:
@tomkrouper the puppet4/strict failure is unrelated to your change
however, i would still appreciate a [proper rebase](https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing)

tomkrouper:
@igalic,

This is my first time following a rebase workflow. Please let me know if I fixed this correctly. Thanks.

igalic:
thanks @tomkrouper! this looks great!

PR #814, opened 2016-03-25
percona 5.7 doesn't have column PASSWORD in mysql.user table

pondohva:
percona 5.7 doesn't have PASSWORD column in mysql.user table neither and have been forked form 5.7.10 mysql, so, I guess, mysql_user provider should act the same way.


igalic:
i have ***no idea*** why this failing on puppet 4 in strict mode.

pondohva:
Hello. Any news? Should I create jira task or smth?

solomonty:
Please take a look at PR #818. I believe it includes your change in addition to other changes needed for Percona.

pondohva:
@solomonty yes it is, thank you. I'll wait for merge of #818 PR

PR #815, opened 2016-03-25
Remove mysql_table_exists() function

hunner:
This function is intended to check for the existence of a table before
declaring some resource, but this is neither portable (because functions
run on the master, not the agent) nor one-run idempotent (because the
function would run before mysql is even installed, and would take two
runs to do anything).

The correct way of doing this would be to update the providers and
dependency ordering to handle the conditional states.

Luckily this was never released so it is backwards compatible.


hunner:
@igalic You merged #776 and #777, so it would be great if you could review and merged this. Thanks

hunner:
Errors unrelated; https://github.com/puppetlabs/puppetlabs-mysql/pull/816 fixes them.

igalic:
*reviews*

cropalato:
@hunner / @igalic 

Can you please let me know what if wrong with that fix?
In my case, I want to be able to skip the grant if the table doesn't exist.
I made my own fix before find that one. But in my case I added a param to decide if I should skip grant command (and show a warning) if skip_missing_table => true (default is false).

Please help me to understand what is wrong and I can address the problem properly.

PR #824, opened 2016-04-14
Default mysqld_type should be "mysql"

ih84ds:
Default mysqld_type return value should be "mysql" if another type is not detected. Returning nil breaks mysql 5.7.11 on Ubuntu (at least) due to the conditional used in mysql_user provider.

### DIFF
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -24,7 +24,7 @@ class Puppet::Provider::Mysql < Puppet::Provider
     mysqld_version_string.scan(/mariadb/i) { return "mariadb" }
     mysqld_version_string.scan(/\s\(mysql/i) { return "mysql" }
     mysqld_version_string.scan(/\s\(percona/i) { return "percona" }
-    nil
+    return "mysql"
   end
 
   def mysqld_type


solomonty:
What is the version string for mysql on Ubuntu?

ih84ds:
mysqld  Ver 5.7.11-0ubuntu6 for Linux on x86_64 ((Ubuntu))

lib/puppet/provider/mysql.rb
```diff
@@ -24,7 +24,7 @@ def self.mysqld_type
     mysqld_version_string.scan(/mariadb/i) { return "mariadb" }
     mysqld_version_string.scan(/\s\(mysql/i) { return "mysql" }
     mysqld_version_string.scan(/\s\(percona/i) { return "percona" }
-    nil
+    return "mysql"
```
solomonty: It is better to match on what we know instead of assuming everything is "mysql"

add this above the `nil` instead

```
mysqld_version_string.scan(/\s\(\(ubuntu/i) { return "mysql" }
```

ih84ds: I disagree. This is the mysql module, after all. mysql should be the default. Why nil? There are Ubuntu packages out there for mariadb as well.

mmailand:
I have the some problem on Ubuntu Xenial.

lib/puppet/provider/mysql.rb
hunner: I believe it's nil specifically for catching unhandled cases instead of propagating incorrectly handled cases. Fail-fast style.

Anyway, I think I agree with @solomonty that if `((Ubuntu))` is the moniker that Ubuntu packagers have chosen to mark their packages with, we can only follow suit.

ih84ds: That sounds OK in theory, but it only works if it is caught where it is used. It is most definitely not caught in the mysql_user provider.

I feel that mariadb and percona are the special cases that need to be trapped, not mysql. If there is a "mysqld" executable, it should be assumed to actually be mysql unless a special case is detected.

hunner: Sounds reasonable.

PR #825, opened 2016-04-19
Additional mysql_datadir exists check for MySQL Enterprise

alwiesner:
MySQL Enterprise rpm's, downloaded directly from Oracle, install an empty 'mysql' dir inside of the datadir (/var/lib/mysql).  Due to this, when the exists? method runs with "File.directory?("#{datadir}/mysql")", since the directory does exist for MySQL Enterprise (though, it is empty), the 'mysql_install_db' script never gets run to install the base mysql DB.

This change adjusts the check to confirm the dir exists and is not empty.


jonnytpuppet:
Closing due to #834 which is a squashed version of this PR)

PR #826, opened 2016-04-20
Master

telmerobert:
update TODO code


hunner:
Thanks for your contribution, but perhaps it is misplaced?

PR #827, opened 2016-04-25
Added bzip2 package support on mysqldump backup

lcrisci:
The script used by this mysqldump class, mysqlbackup.sh, requires bzcat
which is part of the bzip2 package.

This patch will make sure that bzip2 is installed before setting up the
cron tab job using mysqlbackup.sh


vchepkov:
I would use ensure_packages function instead to avoid conflicts with other modules

lcrisci:
I updated my PR to use ensure_packages instead of package.

PR #831, opened 2016-05-05
make sure we find mysqld on FreeBSD

fraenki:
Fixes:

```
Error: Could not prefetch mysql_user provider 'mysql': Command mysqld is missing
```


igalic:
thank you @fraenki!

PR #832, opened 2016-05-06
Find MySQL 5.5 installation on CentOS

jjagodzinski:
MySQL from CentOS SCL repository is installed in different location. You should also check for binaries in this location.
Without this change puppet returns error like: Could not find a suitable provider for mysql_datadir

### DIFF
--- a/lib/puppet/provider/mysql_datadir/mysql.rb
+++ b/lib/puppet/provider/mysql_datadir/mysql.rb
@@ -57,7 +57,7 @@ Puppet::Type.type(:mysql_datadir).provide(:mysql, :parent => Puppet::Provider::M
 
   def exists?
     datadir = @resource[:datadir]
-    File.directory?("#{datadir}/mysql")
+    (File.directory?("#{datadir}/mysql")) && (Dir.entries("#{datadir}/mysql") - %w{ . .. }).any? 
   end
 
   ##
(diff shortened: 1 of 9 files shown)


lib/puppet/provider/mysql_datadir/mysql.rb
```diff
@@ -6,7 +6,7 @@
   initvars
 
   # Make sure we find mysqld on CentOS
-  ENV['PATH']=ENV['PATH'] + ':/usr/libexec'
+  ENV['PATH']=ENV['PATH'] + ':/opt/rh/mysql55/root/usr/bin:/opt/rh/mysql55/root/usr/libexec:/usr/libexec'
```
igalic: how does this work on *BSD?

jjagodzinski: I didn't test it on *BSD. I don't have *BSD in my environment.

jonnytpuppet:
I like the fact that this issue has a "simple" solution, although it is not elegant to pollute an environment's PATH with paths that clearly won't exist.

Keen to hear from @hunner if you can think of a nearly way to do this?

Thanks!

lib/puppet/provider/mysql_datadir/mysql.rb
daenney: How does this commit warrant asking about BSD? It adds a path extremely specific to Red Hat Collections, `/opt/rh`.

hunner:
Could you rebase because we merged #838? And also, would it work to put the custom paths AFTER /usr/libexec instead of before? To limit side-effects of this change on existing installs.

jjagodzinski:
New change in #842 (I've got some problems with rebase).

PR #833, opened 2016-05-06
move out $options['mysqld']['log-error'] from service.pp into installdb.pp

ndelic0:
Installation of Percona-Server-57 from yum repository has a side effect - touching of /var/log/mysqld.log. The file is empty and owned by root user, and that prevents mysql datadir initialization. 

puppet-agent[2003]: Execution of '/usr/sbin/mysqld --initialize-insecure --basedir=/usr --datadir=/var/lib/mysql --user=mysql --log-error=/var/lib/mysql/mysqld.log' returned 1
.
Given issue reproduced on CentOS 6.7, possibly on CentOS7.

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -1,5 +1,6 @@
 #
 class mysql::server::installdb {
+  $options = $mysql::server::options
 
   if $mysql::server::package_manage {
 
@@ -15,6 +16,15 @@ class mysql::server::installdb {
     } else {
       $_config_file=undef
     }
+
+  if $options['mysqld']['log-error'] {
+    file { $options['mysqld']['log-error']:
+      ensure => present,
+      owner  => $mysqluser,
+      group  => $::mysql::server::mysql_group,
+      before => Mysql_datadir[ $datadir ],
+    }
+  }
 
     mysql_datadir { $datadir:
       ensure              => 'present',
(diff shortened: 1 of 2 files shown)


jonnytpuppet:
Hey There

Thanks for your contribution!

Are you sure this isn't more of a bug/issue with Percona-Server-57? Can you please explain how moving the resource from one pp to the other solves this?

Thanks!

ndelic0:
Hi jonnytpuppet!

It's a bug with a yum installer. By managing log_error file inside of installdb class we're only changing file ownership, prior to mysql_datadir resource declaration, thus ensuring that mysql --initialize-insecure actually succeed. 
The way I see, though it is a bug with the installer, we're not modifying module logic here, just securing it.

jonnytpuppet:
Hi There

This is not the correct place to move the file resource. It may fix your solution, however then we run the risk of the file not being managed properly.

What we should do, it make the datadir resource depend on the logfile resource (inside the service class).

Thanks

ndelic0:
Hi JT,

Not sure I got this one. If you're suggesting adding dependency for logfile resource inside service class, basically something like this 

`if $options['mysqld']['log-error'] {
    file { $options['mysqld']['log-error']:
      ensure => present,
      owner  => $mysqluser,
      group  => $::mysql::server::mysql_group,
      before => Mysql_datadir[ $options['mysqld']['datadir'] ],
    }`

smells like a dependency cycle here  

  `Class['mysql::server::installdb'] ->
  Class['mysql::server::service'] ->`

PR #840, opened 2016-05-20
Fix global parameter usage in backup script

HT43-bqxFqB:
If mysqldump uses --defaults-extra-file mysql should use the same. Otherwise settings like different socket paths etc. will not work. Preferably this should be detected via test cases, as backups are quite important...

### DIFF
--- a/templates/mysqlbackup.sh.erb
+++ b/templates/mysqlbackup.sh.erb
@@ -68,7 +68,7 @@ cleanup
 <% end -%>
 <% if @backupdatabases.empty? -%>
 <% if @file_per_database -%>
-mysql --defaults-file=$TMPFILE -s -r -N -e 'SHOW DATABASES' | while read dbname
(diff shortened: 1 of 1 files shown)


DavidS:
There are already a few tests exercising the back up script at https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/acceptance/mysql_backup_spec.rb#L48-L75 why did they miss this issue?

HT43-bqxFqB:
I think those test don't look for any really unusual cases. In our case the mysql instance was under a different mountpoint. This included the socket file. In https://github.com/puppetlabs/puppetlabs-mysql/blob/master/templates/mysqlbackup.sh.erb#L71 the script fails because it can't connect to the database because it uses the default socket file path.
Anyway, I think if the script uses --defaults-extra-file in the mysqldump part, it should use the same logic when trying to connect to mysql to find out what to dump.

igalic:
how about we merge this for now?

PR #841, opened 2016-05-23
Use mariadb by default for Debian Jessie

lomignet:
Just updates params.pp to use mariadb specific packages. All other configurations (files, paths, groups) stay the same.


DavidS:
Closing in favor of #845 , which is scheduled to be merged

PR #843, opened 2016-05-23
parametrize xtradb packagename

ndelic0:
Had a problem with previous pull request.


DavidS:
Thanks for re-submitting.

* under which circumstances is the package name different? Can this be mapped in the params class instead?
* this is a private class, how do you expect people to pass in the value? hiera?

ndelic0:
@DavidS Thanks for pointing that out, believe it's ok now? It appears that default percona-xtradb is not compatible with newer versions of Percona server. https://bugs.launchpad.net/percona-xtrabackup/+bug/1568049

DavidS:
That linked bug is speaking about a requirement to have a new enough version of Percona Xtrabackup to work with 5.7. Now I'm even more confused. What would you put in as package name to address that?

ndelic0:
percona-xtrabackup-2{1..4}

ndelic0:
@DavidS Anything I need to work on here?

tphoney:
@ndelic0 it looks good from me code wise, it would be great if you could squash the commits in the mean time. @DavidS comment still holds, if you look at the reference you can specify a version with ensure: https://docs.puppet.com/puppet/latest/reference/types/package.html#package-attribute-ensure

PR #844, opened 2016-05-24
Fix error in server/installdb: Error: Could not retrieve catalog from…

tdm4:
… remote server: Error 400 on SERVER: no parameter named 'log_error' at installdb.pp:29 on Mysql_datadir[/var/lib/mysql] at installdb.pp:29

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -31,7 +31,6 @@ class mysql::server::installdb {
       datadir             => $datadir,
       basedir             => $basedir,
       user                => $mysqluser,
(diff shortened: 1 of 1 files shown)


DavidS:
But there is? https://github.com/puppetlabs/puppetlabs-mysql/blob/master/lib/puppet/type/mysql_datadir.rb#L30 

Maybe the error is caused by not passing in a value in $mysql::server::options['mysqld']['log-error'], or an invalid one?

tdm4:
I'm going to close this PR.

PR #852, opened 2016-06-03
[MODULES-3441] Discover mysql version using facts

jtopper:
Per https://tickets.puppetlabs.com/browse/MODULES-3441, the mysql module has behaviour which varies by server version. The version is discovered by running `mysqld -V`. On hosts without a MySQL server package install, this fails, which means that contrary to the README, it's not actually possible to use this module to manage a remote db.

This PR moves the version string discovery into a new fact, `mysqld_version` which is used by the provider.  This makes it possible to configure the db version with a custom fact when a remote db (eg AWS RDS) is being managed.

### DIFF
--- a/lib/facter/mysqld_version.rb
+++ b/lib/facter/mysqld_version.rb
@@ -1,0 +1,5 @@
+Facter.add("mysqld_version") do
+  setcode do
+    Facter::Util::Resolution.exec('mysqld -V')
+  end
+end
--- a/README.md
+++ b/README.md
@@ -152,6 +152,8 @@ host=localhost
 password=secret
 ```
 
+This module uses the `mysqld_version` fact to discover the server version being used.  By default, this is set to the output of `mysqld -V`.  If you're working with a remote MySQL server, you may need to set a custom fact for `mysqld_version` to ensure correct behaviour.
+
 When working with a remote server, do *not* use the `mysql::server` class in your Puppet manifests.
 
 ### Specify passwords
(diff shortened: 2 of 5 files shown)


lib/facter/mysqld_version.rb
```diff
@@ -0,0 +1,5 @@
+Facter.add("mysqld_version") do
+  setcode do
+    Facter::Util::Resolution.exec('mysqld -V')
```
daenney: Does this one not need the `.compact` method chained on it, like it used to do on the provider?

jtopper: I think the `.compact` is superfluous.  It'd remove nil elements from a list.  In the provider, the call to `.compact` is run against an anonymous, single-item list, where it will have no effect.

daenney: 💯

bmjen:
Thanks @jtopper, this looks good and satisfies a need. Can you squash the commits?

jtopper:
I can squash the commits by all means, but the contributing doc requests that contributors "Make commits of logical units".  Which is it? :)

daenney:
I would squash the docs into the previous commit, since using the fact and updating the documentation about the fact that this is the case should be one thing.

bmjen:
Yep. What @daenney said. We usually like to combine feature+test+docs into 1 logical commit. So if a reversion or cherry-pick needed to happen, it all comes out as one logical chunk.

jtopper:
Cool, no problem. I've just squashed everything.

DavidS:
Thanks for your contribution!

daenney:
Woohoo 😄.

PR #853, opened 2016-06-07
Resource fails when fqdn is not set.

ragonlan:
if server don't have domain name facter don't initialize fqdn and mysql_user resource fails with "...Invalid database user..." error.

### DIFF
--- a/manifests/server/account_security.pp
+++ b/manifests/server/account_security.pp
@@ -16,7 +16,7 @@ class mysql::server::account_security {
       require => Anchor['mysql::server::end'],
     }
   }
-  if ($::fqdn != 'localhost') {
+  if ($::fqdn and $::fqdn != 'localhost') {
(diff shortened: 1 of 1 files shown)


igalic:
@ragonlan i'm confused. why would your servers *not* have an fqdn??

ragonlan:
Sorry it was a bug in old facter versions (previous to 2.1.0-1puppetlabs1) fqdn is not defined unless host and domain is defined:
```
    if host and domain
      [host, domain].join(".")
    else
      nil
    end
```
 it is fixed in current version with:

```
     if host and domain
       [host, domain].join(".")
     elsif host
       host
     else
       nil
     end
```

but official version from ubuntu 14.04 (facter=1.7.5-1ubuntu1) is affected.

igalic:
thank you very much @ragonlan!

PR #857, opened 2016-06-16
Add example of MariaDB server installation on Ubuntu

aerostitch:
Hi,

As I've seen quite often questions on how you install a MariaDB server on Ubuntu, I decided to add an example to the documentation.
I've tested this minimal example on an Ubuntu Trusty server and it worked.

Thanks for the reviews.
Joseph


aerostitch:
Added a reference to that PR in the official bug tracker: https://tickets.puppetlabs.com/browse/MODULES-3479

bmjen:
Thanks for the contribution @aerostitch !

PR #863, opened 2016-07-08
Fixing server id 32 bit overrun error

NoodlesNZ:
As per comments on this commit https://github.com/puppetlabs/puppetlabs-mysql/commit/96c9d6d9a749e0e68f6006c80c86fcfab72d9763 we need to roll this back.

If the original committer still has issues then we should find a better solution


benohara:
Think I reported the original issue, we still use a mix of the 'fix' and custom code (based of location_id and ip) - seen duplicates with this also.

Can you take the last 4 chunks of the mac for the calculation? should be more 'unique' missing out more of the manufacturer bit of it?

tphoney:
It looks like #872 fixes this issue. Thanks for looking into this problem.

PR #864, opened 2016-07-08
Fixing logdir dependency issue

NoodlesNZ:
mysql user doesn't exist until mysql-server package is installed

I didn't notice this problem until trying to install a server from scratch.


DavidS:
A fix for this has been merged in #868

PR #866, opened 2016-07-09
don't manage database resources if service is disabled

vchepkov:
currently module fails if service_enabled is set to false, since it's trying to manage database objects, which are not available


eputnam:
@vchepkov , thanks for the PR. Due to inactivity, I'm going to close this. If this is still something you'd like to see merged, please feel free to re-open!

PR #867, opened 2016-07-14
MODULES-3601 Change install order.

QuentinMoss:
Fresh installation of Percona 56 fails due to the order of server installation / configuration. The module first tries to create the data directory and `chown` to the `mysql` system user, but this user does not exist until the package is fully installed. 

https://tickets.puppetlabs.com/browse/MODULES-3601

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -75,8 +75,8 @@ class mysql::server (
   }
 
   Anchor['mysql::server::start'] ->
-  Class['mysql::server::config'] ->
   Class['mysql::server::install'] ->
+  Class['mysql::server::config'] ->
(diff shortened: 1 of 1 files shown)


hunner:
This was caused by https://github.com/puppetlabs/puppetlabs-mysql/pull/813 and was to solve another issue, so simply reverting the change is not going to solve the problem. I think installation/configuration steps are more involved in some edge cases such as you mention in the ticket when logbin's are created.

Perhaps we could just move the logbindir logic to a later stage in the installation?

QuentinMoss:
@hunner As long as logbindir happens after install all is fine! Thanks for reviewing.

PR #868, opened 2016-07-20
MODULES-3601 Move binary logging configuration to take place after pa…

QuentinMoss:
…ckage install

https://tickets.puppetlabs.com/browse/MODULES-3601

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -55,6 +55,7 @@ class mysql::server (
 
   include '::mysql::server::config'
   include '::mysql::server::install'
+  include '::mysql::server::binarylog'
   include '::mysql::server::installdb'
   include '::mysql::server::service'
   include '::mysql::server::root_password'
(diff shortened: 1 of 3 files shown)


QuentinMoss:
@hunner Sorry for closing the previous PR. Shortly after I thought I would give this a try. Let me know what you think.

eborned:
Wouldn't this still result in errors when the binary log directory is unavailable during setup? The installation of the mysql-server package normally auto-starts the server. The config file has been setup to work with the log-bin directory, but that directory is created after the service has been started.

QuentinMoss:
@eborned Here is a test deployment on CentOS 7 with binary log options defined:

https://gist.github.com/QuentinMoss/14f560dfc7cac76282639a66462f0392

hunner:
I think just installing the package on debian/ubuntu auto-starts the service, but this will at least fix the currrent issue and allow it to work on redhat.

bwitt:
Yes installing the package on ubuntu auto-starts the service, which results in a failure.

PR #869, opened 2016-07-21
(MODULES-3401) Fix for mysql version retrieval

HelenCampbell:


HelenCampbell:
Fixes issues around the retrieval of mysql version string, though further issues still remain these will be addressed in a separate issue.

HelenCampbell:
Hey @jtopper! When does this code run on a host that neither has mysqld or the fact set? Without that information, the provider cannot generate the correct SQL and interactions, and will only fail later when trying to talk - wrongly - to the server, which is why this fix was implemented. Does that help clarify for you?

jtopper:
@HelenCampbell We manage Amazon RDS MySQL instances from hosts that don't have mysqld installed.  In a number of our use cases, the default queries that the module builds are sufficient, so we haven't introduced facts (though obviously we will now be doing so, on account of this change :)

Ultimately I think you're right that this is an unusual case (this module really isn't great at managing remote databases), but I think the flow here of "look up the fact, then call mysqld -V" is a bit weird when the fact itself falls back to calling mysqld -V.  Perhaps it should throw a more obvious error, with instructions on adding an appropriate version fact?

PR #871, opened 2016-08-04
MODULES-3697 Changed puppet fail behaviour for mysql create user and grant if user name is longer than 16 chars

dn1s:
User name length has changed "MySQL user names can be up to 32 characters long (16 characters before MySQL 5.7.8)"
see http://dev.mysql.com/doc/refman/5.7/en/user-names.html

Tested on CentOS release 6.8 (Final) with mysql version 5.7.14


dn1s:
Tested on CentOS release 6.8 (Final) with mysql version 5.7.14

bmjen:
Thanks @dn1s !

PR #875, opened 2016-08-13
 MODULES-3729 - puppetlabs/MySQL module -- Percona/MySQL 5.7 initialize bug

QuentinMoss:
There is a bug in the Puppetlabs/MySQL module that affects MySQL/Percona 5.7. Please see below:

```ruby
  def exists?
    datadir = @resource[:datadir]
    (File.directory?("#{datadir}/mysql")) && (Dir.entries("#{datadir}/mysql") - %w{ . .. }).any?
  end
```

This code appends 'mysql' to the specified datadir, for example, if we set a datadir to /var/lib/mysql, it will look for /var/lib/mysql/mysql, which never exists and initialize is always called. When called, it fails because the data directory exists.

See bug report here: https://tickets.puppetlabs.com/browse/MODULES-3729

### DIFF
--- a/lib/puppet/provider/mysql_datadir/mysql.rb
+++ b/lib/puppet/provider/mysql_datadir/mysql.rb
@@ -57,7 +57,7 @@ Puppet::Type.type(:mysql_datadir).provide(:mysql, :parent => Puppet::Provider::M
 
   def exists?
     datadir = @resource[:datadir]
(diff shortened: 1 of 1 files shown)


QuentinMoss:
@bmjen You can close this PR - the existing code is functioning properly.

bmjen:
Sounds good @QuentinMoss . Thanks!

PR #877, opened 2016-08-17
Ensure that error log is writable by owner

runejuhl:
In the case that the Puppet manifest has

```puppet
File {
  mode => '0444',
}
```
this will be inherited and render the log file unwritable, unless we specify mode.


bmjen:
Thanks for the contribution @runejuhl !

PR #883, opened 2016-09-03
Silence STDERR in Facter

dennisdegreef:
Currenly performing a `puppet apply`, the following output is shown

```
Info: Loading facts
160903 16:55:09 [Note] /usr/sbin/mysqld (mysqld 10.0.26-MariaDB-0+deb8u1) starting as process 6358 ...
Notice: Compiled catalog for foo.example.com in environment production in 1.52 seconds
```

This is due to Facter performing `mysqld -V` to determine the version as fact.

The output of this command alone is this:
```
$ sudo mysqld -V
mysqld  Ver 10.0.26-MariaDB-0+deb8u1 for debian-linux-gnu on x86_64 ((Debian))
160903 17:02:08 [Note] mysqld (mysqld 10.0.26-MariaDB-0+deb8u1) starting as process 7260 ...
```
Where the line with 'starting as process' is STDERR.

This diff redirects STDERR to /dev/null, since it probably contain any critical information (during a version check). 

After applying this diff, the expected outcome is the same again
```
$ sudo mysqld -V 2>/dev/null
mysqld  Ver 10.0.26-MariaDB-0+deb8u1 for debian-linux-gnu on x86_64 ((Debian))
```

As well as within Puppet.

```
Info: Loading facts
Notice: Compiled catalog for foo.example.com in environment production in 1.52 seconds
Info: Applying configuration version '1472915111'
```

### DIFF
--- a/lib/facter/mysqld_version.rb
+++ b/lib/facter/mysqld_version.rb
@@ -1,5 +1,5 @@
 Facter.add("mysqld_version") do
   setcode do
-    Facter::Util::Resolution.exec('mysqld -V')
+    Facter::Util::Resolution.exec('mysqld -V 2>/dev/null')
   end
 end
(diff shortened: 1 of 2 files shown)


eputnam:
Thanks for the PR!

PR #884, opened 2016-09-06
Change the order of the config and install

vjekoslav:
Currently default config file gets overwritten by installation files, and it even fails if there is no supported directory structure for the file.

In my case, it was MySQL community edition 5.7 on Ubuntu 14.04.


igalic:
maybe we should be smarter about this, this time around https://github.com/puppetlabs/puppetlabs-mysql/pull/813

bmjen:
Agree with @igalic.

rnelson0:
@igalic In reference to #813, I think any concern about going back to the pre-813 ordering would be around improperly designed packages. For instance, at http://fedoraproject.org/wiki/EPEL:SysVInitScript#Initscript_packaging under "Why don't we...." mentions that services should not be started by the package installation due to a number of valid reasons. With mariadb-server-5.5 on EL7 I do not see such post-install scripts. Of course, we can't verify all packages are built right, and we need testing to ensure this fix works with properly packaged scripts but I think this PR is appropriate in its intentions.

Poil:
For Debian family, I've made a script which remove the service start via and apt rules ```DPkg::Pre-Install-Pkgs {"xargs -rL1 /usr/local/bin/stripdeb.sh 2>&1 | logger -t stripdeb"}```

But I think, the puppet module must manage the mysql process user and we must create the binlogdir before the package installation.

eputnam:
Thanks @vjekoslav! This is something we've gone back and forth on many times and it's going to need a deeper change. Please feel free to re-open this PR with some more details on the problem and the proposed solution. @Poil could you talk a little more about your thoughts on this in a new PR?

PR #890, opened 2016-09-20
(MODULES-3623) Centralise MySQL calls, so they can all be modified fr…

wayland:
…om one place

The current codebase handles each mysql call via calls to the command-line
mysql client.  There are basically only two sets of options passed, but the
options are repeated separately in various places throughout the code.  The
new self.mysql function now issues the call to the mysql client with one of
the two sets of options (selected via an optional function parameter).

Now the options to all mysql calls can be modified from one place.  This will
be useful when we are converting the module to allow for databases on multiple
remote servers, since we will be able to add the -h switch to all calls at
once.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -3,3 +3,22 @@ Gemfile:
   optional:
     ':development, :unit_tests':
       - gem: rspec-puppet-facts
+.travis.yml:
+  includes:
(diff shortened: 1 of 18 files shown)


hunner:
Closing in favor of #904

PR #894, opened 2016-09-29
MODULES-3904: Abiliity to use SQL_LOG_BIN for MySQL types

tomkrouper:
There are times where you may want a `mysql_user{}` or mysql_grant{}`to
exist only on one host. The addition of session_sql_log_bin can be set
to enable or disable binary logging for the change. This also works for
`mysql_database{}` but I believe this would be used less often.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,6 +473,7 @@ users => {
     max_updates_per_hour     => '0',
     max_user_connections     => '0',
     password_hash            => '*F3A2A51A9B0F2BE2468926B4132313728C250DBF',
+    session_sql_log_bin      => '1',
   },
 }
 ```
@@ -484,11 +485,12 @@ Optional hash of grants, which are passed to [mysql_grant](#mysql_grant).
 ```
 grants => {
   'someuser@localhost/somedb.*' => {
-    ensure     => 'present',
-    options    => ['GRANT'],
-    privileges => ['SELECT', 'INSERT', 'UPDATE', 'DELETE'],
-    table      => 'somedb.*',
-    user       => 'someuser@localhost',
+    ensure              => 'present',
+    options             => ['GRANT'],
+    privileges          => ['SELECT', 'INSERT', 'UPDATE', 'DELETE'],
+    table               => 'somedb.*',
+    user                => 'someuser@localhost',
+    session_sql_log_bin => '1',
   },
 }
 ```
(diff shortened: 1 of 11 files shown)


rfkrocktk:
@tomkrouper This is something where you write conditional logic. Define a custom Ruby fact that determines whether a given database is the primary/master, and then write Puppet code that exploits that:

```
class roles::mysql::users {
    if $::is_mysql_master { 
        mysql_user { ...:
            ...
        }
    }
}
```

tomkrouper:
@rfkrocktk, That's how we currently have our system setup. However, I've worked at previous companies where we only wanted to have users on certain servers. For example, individual accounts on an analytics replica. We didn't want the user on the master or production read replicas. Currently I am wanting puppet to not error out on setting up external scripts because the users is not created until it is in the replication chain.

rfkrocktk:
@tomkrouper Understood. I am not the maintainer of this repository, but my guess is that this is a very rare use-case and might not be supported. Mucking with the binary log is pretty risky business. If I had to do this, I'd probably prefer something like this:

```
mysql_database { 'mydatabase':
     manageresource => $::is_mysql_master
}

mysql_user { 'user@mydatabase':
    ...
}

Mysql_database['mydatabase'] -> Mysql_user['user@mydatabase']
```

The logic here would be for puppet to not manage the actual database and simply wait for it to appear possibly with a given timeout. 

In either case, it seems like this is a pretty rare use-case. I'd use LDAP for managing these kind of constraints.

wilson208:
@tomkrouper I am closing this PR because it hasnt seen any activity recently. If you want to work on this again don't hesitate to re-open!

tomkrouper:
@wilson208, when I first created, I believe I had it working. I got the impression no one was interested in it being an option. If someone else finds the change valuable, I can rebase and retest. Otherwise, it is fine to leave closed.

elfchief:
I would really _really_ like to have this feature available. Is this something that we could potentially reopen and get merged? I realize that there are multiple schools of thought about whether things like users and grants should be replicated or not, but "don't replicate" is definitely a valid perspective and outright required under some circumstances. (And just adding a replication filter for the 'mysql' table is insufficient in some circumstances -- such as when using gtid)

ParshinPavel:
Hi there,

It is really useful PR which helps to prevent occurring errant GTID on replicas in the cluster. Just now we have to manually fix them every time when we change user grants only on replicas.

gdubicki:
Please see #1355 that implements this on top of the currently latest version of the module.

PR #896, opened 2016-10-04
Add support for setting tls options for mysql_user's

JAORMX:
This adds support for adding or modifying tls options via the
mysql_user provider. this works with newer mysql syntax which uses
the ALTER USER syntax and also with the older syntax which relies on
using GRANTs.


JAORMX:
any more feedback for this?

PR #902, opened 2016-10-26
Enhancements to xtrabackup backup provider

fraenki:
This PR will bring the `xtrabackup` provider up to par with the `mysqldump` provider and adds one other enhancement:
- Add support for existing parameters `$backupuser` and `$backuppassword`
- Automatically create backup user/grants
- Add support for the `databases` parameter to provide a list of databases to back up
- Add support for optional arguments to `innobackupex` by adding the new parameter `$optional_args`

This makes it possible to specify additional arguments to be passed to `innobackupex`:

```
mysql::server::backup:
  provider: 'xtrabackup'
  user: 'backupusr'
  password: 'secret'
  databases:
    - 'foo'
    - 'bar'
  optional_args:
    - '--no-lock'
    - '--parallel=2'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,10 @@ Sets the server backup implementation. Valid values are:
 
 Defines the maximum SQL statement size for the backup dump script. The default value is 1MB, as this is the default MySQL Server value.
 
+##### `optional_args`
+
+Specifies an array of optional arguments which should be passed through to the backup tool. (Currently only supported by the xtrabackup provider.)
+
 #### mysql::server::monitor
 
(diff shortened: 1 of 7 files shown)


bmjen:
thanks @fraenki can you take a look at the spec test failures? Also, the new options need to be documented in the README.

fraenki:
@bmjen I've fixed the unit test and added documentation for the only _new_ parameter `optional_args` (all other options aren't new, but were not supported previously by the xtrabackup provider).

bmjen:
Thanks @fraenki !

PR #904, opened 2016-11-02
(MODULES-3623) Centralise MySQL calls...

wayland:
...so they can all be modified from one place.

The current codebase handles each mysql call via calls to the command-line
mysql client.  There are basically only two sets of options passed, but the
options are repeated separately in various places throughout the code.  The
new self.mysql function now issues the call to the mysql client with one of
the two sets of options (selected via an optional function parameter).

Now the options to all mysql calls can be modified from one place.  This will
be useful when we are converting the module to allow for databases on multiple
remote servers, since we will be able to add the -h switch to all calls at
once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1048,3 +1048,4 @@ This module is based on work by David Schmitt. The following contributors have c
 * Chris Weyl
 * Daniël van Eeden
 * Jan-Otto Kröpke
+* Timothy Sven Nelson
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -57,13 +57,15 @@ class Puppet::Provider::Mysql < Puppet::Provider
   def self.mysql(textOfSQL, type)
     if type.eql? 'system'
       mysql([defaults_file, '--host=', system_database, '-e', textOfSQL].flatten.compact)
-    else
+    elsif type.eql? 'regular'
       mysql([defaults_file, '-NBe', textOfSQL].flatten.compact)
+    else
+      raise Puppet::Error, "#mysql had an error -> Unrecognised type '#type'"
     end
   end
 
   def self.users
-    self.mysql("SELECT CONCAT(User, '@',Host) AS User FROM mysql.user").split("\n")
+    self.mysql("SELECT CONCAT(User, '@',Host) AS User FROM mysql.user", 'regular').split("\n")
   end
 
   # Optional parameter to run a statement on the MySQL system database.
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -54,8 +54,16 @@ class Puppet::Provider::Mysql < Puppet::Provider
     self.class.defaults_file
   end
 
+  def self.mysql(textOfSQL, type)
+    if type.eql? 'system'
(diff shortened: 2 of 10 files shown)


bmjen:
Thanks @wayland . This PR has some failing tests. Also how does this PR connect with #900 and #890? Are they chained? or should those be closed in favor of this one?

wayland:
Yes, this should take precedence.  There were some git conflicts, and I couldn't figure any better way to resolve them.  With the first one, I ran as many tests on my local machine as I could (without installing Docker, etc), and then let the system here catch the remaining problems.  I got a bit slacker in the later rounds.  I agree the test is having problems, but I'm not sure I understand the testing system well enough to resolve this one.  I'll keep looking at it for a while, and let you know if I'm really stuck.

wilson208:
@wayland are you able to reproduce the fails on your own machine?

wayland:
Unfortunately not; that's why I keep committing bad code.  Those tests seem to require vagrant, which I don't have installed.  If you have any thoughts about what the problem might be, that'd be helpful (note: I just changed the mysql() function call to have 2 parameters instead of 1, and the error may somehow be related).

wilson208:
Without going in and cloning your fork and debugging the tests themselves I couldn't say. However I couldn't consider merging this PR unless the acceptance tests are passing.

If you don't wish to install vagrant, these tests can also be ran using docker. You will need to install docker on your machine.
`PUPPET_INSTALL_TYPE=agent BEAKER_set=docker/centos-7 BEAKER_debug=true BEAKER_destroy=no BEAKER_provision=yes bundle exec rspec spec/acceptance`

The above will use the centos-7 nodeset inside spec/acceptance/nodesets/docker, however, ensure the acceptance tests pass on all 3 docker nodesets (centos-7, debian-8 & ubuntu-14.04).

bmjen:
@wayland the change to mysql() is exactly why the acceptance tests are failing. It appears that the acceptance tests need to be updated to pass in 2 parameters, unless you are able to provide a reasonable default for the new parameter such as `regular`.

This is the error from all of the acceptance tests:

```
Error: Failed to apply catalog: wrong number of arguments (1 for 2)
```

hunner:
Thanks for opening this issue. It hasn't seen activity in a while, so I'm going to close the issue. This doesn't mean that we don't think this issue is valid, we're just trying to assess what issues are still relevant. If you still need this(even if you're not willing to work on it yourself) please feel free to reopen it. Again thanks for opening the issue, users who report bugs and ask for features are some of the most important players in the open source community. Feel free to hop by #puppet-dev on irc.freenode.net if you'd like to chat with us about any of this.

hunner:
Actually never mind, this is cool. I'll try and figure it out.

wayland:
@Wilson: Thanks for your Docker tips; I can now run the tests myself.  I also agree that it shouldn't be merged until it works.  
@bmjen: Thanks for your help!  I've made a number of changes, and eliminated the error you mentioned.  Is there any chance you could have another look and see if you can offer me tips on my two current issues?  
@hunner: Thanks for keeping this open.

eputnam:
@hunner , have you had a chance to look at this yet?

hunner:
@eputnam not yet. I still have it open in my "todo tabs list" though.

hunner:
Merged in #1036 \o/

PR #912, opened 2016-11-16
(MODULES-4115) Invalid parameter provider on Mysql_user[user@localhost] in mysql::db

ryanb-hc:
remove provider parameter from $user_resource in mysql::db


hunner:
This is actually a bug in puppet: https://tickets.puppetlabs.com/browse/PUP-1515

PR #917, opened 2016-12-07
Allow mysqltuner to be downloaded via a proxy

Boran:
Extend mysql::server::mysqltuner with an $environment, which is then passed onto staging::file, so that a proxy variable can be passed that allows mysqltuner to be downloaded. Example usage.
```
  class { 'mysql::server::mysqltuner':
    environment => 'https_proxy=http://myproxy.example.ch:80',    # for download
  }
```

### DIFF
--- a/manifests/server/mysqltuner.pp
+++ b/manifests/server/mysqltuner.pp
@@ -37,7 +37,7 @@ class mysql::server::mysqltuner(
     }
 
     staging::file { "mysqltuner-${_version}":
-      source => $_source,
+      source      => $_source,
       environment => $environment,
     }
     file { '/usr/local/bin/mysqltuner':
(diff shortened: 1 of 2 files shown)


manifests/server/mysqltuner.pp
```diff
@@ -37,6 +38,7 @@
 
     staging::file { "mysqltuner-${_version}":
       source => $_source,
+      environment => $environment,
```
bmjen: Indentation needs to be fixed for this in order to pass lint.

bmjen:
@Boran Thanks for the PR! The new parameter needs to be added to the README. Also puppet-lint check is failing and needs to be addressed.

tphoney:
Thanks for the PR, this seems to be a common issue.

PR #920, opened 2017-01-09
Use gfind on solaris

marvin0815:
Solaris' find does not support -maxdepth. One need to use gfind to make this work.

Kind Regards
Oliver


wilson208:
LGTM 👍

PR #921, opened 2017-01-12
Allow custom type providers to be noop

dprince:
I was trying to run a MySQL catalog in a docker container
to generate config files only like this:

 puppet apply --tags 'file,file_line,concat' ...path to manifest....

I was surprised that --tags didn't exclude the custom mysql_ resources
(mysql_user, mysql_database, etc.). As a work around to that I found
that I can implement a noop shim in the Mysql module itself and
then append this to my catalog on the fly to successfully generate
config files within a docker container:

  Mysql_datadir <| |> { provider => noop}
  Mysql_user <| |> { provider => noop }
  Mysql_database <| |> { provider => noop }


dprince:
I realize this needs unit test coverage but please comment on the approach first. Am I missing something here regarding the --tags feature and custom providers?

dprince:
Closing in favor of this: https://review.openstack.org/#/c/423571/

PR #924, opened 2017-01-26
(MODULES-4340) Fix/update Arch Linux paths

m01:
Without this, the puppet MySQL classes cannot interact with MySQL when
the manage_config_file and purge_conf_dir settings are false.
This commit fixes the issue by bringing the defaults in the puppet code
in line with the Arch Linux mariadb package.


m01:
Looks like another change is required for this to not affect setups where Puppet is managing the `my.cnf` config file..

PR #925, opened 2017-02-06
MODULES-2675 - puppetlabs-mysql server_id is not unique

jmk61:
My previous PR which was accepted in 3.7 has been known to cause an error (a unique value higher than the allowed limit of 4294967295. I have modified my facter code to key off of the IP address of the machine.


jmk61:
I don't think the CI checks failed because of the PR...It is throwing Hiera issues.

spec/unit/facter/mysql_server_id_spec.rb
```diff
@@ -6,21 +6,21 @@
...
       it do
-        Facter.fact(:mysql_server_id).value.to_s.should == '4116385'
+        Facter.fact(:mysql_server_id).value.to_s.should == '1739024'
       end
     end
 
     context "node with lo only" do
       before :each do
-        Facter.fact(:macaddress).stubs(:value).returns('00:00:00:00:00:00')
+        Facter.fact(:ipaddress_lo).stubs(:value).returns('172.0.0.1)
```
eputnam: looks like you're missing a single quote here, which is causing the unit tests to choke

jmk61:
@eputnam 
Thanks for the catch. I've fixed the unit test and updated my deprecated items. This is ready for merge.

Thanks
Joey

tphoney:
@jmk61 does the change in https://github.com/puppetlabs/puppetlabs-mysql/commit/8d4b83159e4f8362e1c8e19e80f22079a5a0f547 not resolve the issue ?

jmk61:
@tphoney yes I believe that does fix the issue. Please Close this request

PR #929, opened 2017-03-08
test whether sql paths/import files exist before import

rockpapergoat:
because the sql parameter also takes an array, using an `onlyif` in the
exec isn't appropriate. `test -f` fails with multiple paths passed
to it. this may be dirty but should work. if $sql_paths_exist is false,
don't import. perhaps this should fail instead.

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -22,7 +22,8 @@ define mysql::db (
     fail('$sql must be either a string or an array.')
   }
 
-  $sql_inputs = join([$sql], ' ')
+  $sql_inputs      = join([$sql], ' ')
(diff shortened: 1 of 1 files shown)


wilson208:
@rockpapergoat Thanks for this, however I think it would be better than silently not importing files it would be best to raise a legible error. If you can update this PR to raise the error and also so that the spec and acceptance tests pass I am more than happy to get this merged :)

Thanks!

rockpapergoat:
haha… yes, i realized that *while* submitting the PR. will amend. thanks!

tphoney:
Hi @rockpapergoat , thanks for the work you have put into this pull request. However due to inactivity this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PR #930, opened 2017-03-08
replace validate_* with datatypes in db.pp

bastelfreak:


bastelfreak:
the goal of this is to get rid of all the validate_* calls because they got deprecated. It is not the goal to provide datatypes for all the parameters.

bastelfreak:
ping. Are there any current plans to drop puppet4 on this module?

metadata.json
```diff
@@ -72,7 +72,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 3.0.0 < 5.0.0"
+      "version_requirement": ">= 4.6.1 < 5.0.0"
```
eputnam: our lower bound is 4.7.0

PR #931, opened 2017-03-12
Add example how to install MySQL 5.6 on CentOS 7

nvtkaszpir:
This minimal example shows how to install MySQL Community Server 5.6 on Centos 7.3 using Puppet 3.8.7 using hiera and puppet-mysql=3.9.0

Should clarify why #757 does not really make sense, also should help people with fully working example without needing to read about Percona example ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,7 +324,12 @@ Class['::mysql::client']
 
 ### Install MySQL Community server on CentOS
 
-This minimal example shows how to install MySQL Community Server 5.6 on Centos 7.3 using Puppet 3.8.7 using hiera with puppetlabs-mysql=3.9.0
+You can install MySQL Community Server on CentOS using the mysql module and Hiera. This example was tested with the following versions:
+
+* MySQL Community Server 5.6
+* Centos 7.3
+* Puppet 3.8.7 using hiera
+* puppetlabs-mysql module v3.9.0
 
 In Puppet:
 
--- a/README.md
+++ b/README.md
@@ -322,6 +322,62 @@ Class['apt::update'] ->
 Class['::mysql::client']
 ```
 
+### Install MySQL 5.6 server on CentOS
+
+This minimal example shows how to install MySQL Community Server 5.6 on Centos 7.3 using Puppet 3.8.7 using hiera with puppetlabs-mysql=3.9.0
+
+Puppet:
+
+```puppet
+  include ::mysql::server
+
+  create_resources(yumrepo, hiera('yumrepo', {}))
+
+  Yumrepo['repo.mysql.com'] -> Anchor['mysql::server::start']
+  Yumrepo['repo.mysql.com'] -> Package['mysql_client']
+
+  create_resources(mysql::db, hiera('mysql::server::db', {}))
+```
+
+Hiera entry:
+
+```yaml
+---
+# mysql module requies feeding it a bunch of parameters to properly install mysql, and not mariadb
+# centos 7.3
+yumrepo:
+  'repo.mysql.com':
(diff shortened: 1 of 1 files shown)


tphoney:
@jbondpdx can you give this a once over.

README.md
```diff
@@ -322,6 +322,62 @@ Class['apt::update'] ->
 Class['::mysql::client']
 ```
 
+### Install MySQL 5.6 server on CentOS
```
jbondpdx: Suggest changing to ### Install MySQL on CentOS

README.md
```diff
@@ -322,6 +322,62 @@ Class['apt::update'] ->
...
+
+This minimal example shows how to install MySQL Community Server 5.6 on Centos 7.3 using Puppet 3.8.7 using hiera with puppetlabs-mysql=3.9.0
```
jbondpdx: Unless the version numbers are required (that is, the concept doesn't work with other version numbers), I would remove them from the basic description. Suggest changing to:

This example shows how to install MySQL Community Server on CentOS, using Puppet and Hiera.

README.md
```diff
@@ -322,6 +322,62 @@ Class['apt::update'] ->
...
+
+Puppet:
```
jbondpdx: Suggest

In Puppet:

README.md
```diff
@@ -322,6 +322,62 @@ Class['apt::update'] ->
...
+
+  create_resources(yumrepo, hiera('yumrepo', {}))
+
+  Yumrepo['repo.mysql.com'] -> Anchor['mysql::server::start']
+  Yumrepo['repo.mysql.com'] -> Package['mysql_client']
+
+  create_resources(mysql::db, hiera('mysql::server::db', {}))
+```
+
+Hiera entry:
```
jbondpdx: Suggest:

In Hiera:

README.md
```diff
@@ -322,6 +322,62 @@ Class['apt::update'] ->
...
+
+```yaml
+---
+# mysql module requies feeding it a bunch of parameters to properly install mysql, and not mariadb
```
jbondpdx: typo in "requires"

suggest changing "...and not mariadb" to "instead of mariadb" for clarity

README.md
nvtkaszpir: My idea was to provide real working example of the code, so I've added those versions on purpose, because I did not have time to test it in different setup.
Also yum repo used in hiera explicitly defines MySQL version to use.

I hope this would be a good starting point for anyone else trying to define more generic MySQL section in the future.

README.md
nvtkaszpir: Maybe 'MySQL Community', then?
See below for better explanation.

README.md
jbondpdx: That makes sense. I think the version numbers make the sentence harder to parse, so I'd suggest something like:

You can install MySQL Community Server on CentOS using the mysql module and Hiera. This example was tested with the following versions:

* MySQL Community Server 5.6
* CentOS 7.3
* Puppet 3.8.7
* puppetlabs-mysql 3.9.0

nvtkaszpir: Good idea.
Done.

README.md
nvtkaszpir: Done.

README.md
nvtkaszpir: Done.

README.md
nvtkaszpir: Fixed with other typos.

tphoney:
thanks for all the great work @nvtkaszpir

PR #932, opened 2017-03-14
(#3243) Pass authstring (password_hash) for plugins in mysql_user

jhriggs:
Various authentication plugins use the authentication string to pass additional settings (e.g. `CREATE USER <user>@<host> IDENTIFIED WITH <plugin> AS <plugin_settings>`). The mysql_user provider currently only passes this data for sha256_password. This change passes the authentication string for all plugins when specified.

This should allow, for example:

```puppet
# Percona's auth_pam or auth_pam_compat:
mysql_user { 'foo@localhost':
  ensure        => present,
  password_hash => 'mysql, root=developer, users=data_entry',
  plugin        => 'auth_pam_compat',
  tls_options   => 'SSL',
}

# MySQL Enterprise's authentication_pam:
mysql_user { 'bar@localhost':
  ensure        => present,
  password_hash => 'mysql, root=developer, users=data_entry',
  plugin        => 'authentication_pam',
  tls_options   => 'SSL',
}
```

### DIFF
--- a/lib/puppet/provider/mysql_user/mysql.rb
+++ b/lib/puppet/provider/mysql_user/mysql.rb
@@ -64,7 +64,7 @@ Puppet::Type.type(:mysql_user).provide(:mysql, :parent => Puppet::Provider::Mysq
     # Use CREATE USER to be compatible with NO_AUTO_CREATE_USER sql_mode
     # This is also required if you want to specify a authentication plugin
     if !plugin.nil?
-      if plugin == 'sha256_password' and !password_hash.nil?
+      if !password_hash.nil?
         mysql([defaults_file, system_database, '-e', "CREATE USER '#{merged_name}' IDENTIFIED WITH '#{plugin}' AS '#{password_hash}'"].compact)
       else
         mysql([defaults_file, system_database, '-e', "CREATE USER '#{merged_name}' IDENTIFIED WITH '#{plugin}'"].compact)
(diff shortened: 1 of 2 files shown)


jhriggs:
Any takers on this PR? It is a pretty innocuous change to the code for a much-needed feature.

lib/puppet/provider/mysql_user/mysql.rb
```diff
@@ -123,11 +125,7 @@ def password_hash=(string)
     else
       # Version >= 5.7.6 (many password related changes)
       if (mysqld_type == "mysql" or mysqld_type == "percona") and Puppet::Util::Package.versioncmp(mysqld_version, '5.7.6') >= 0
-        if string.match(/^\*/)
-          mysql([defaults_file, system_database, '-e', "ALTER USER #{merged_name} IDENTIFIED WITH mysql_native_password AS '#{string}'"].compact)
-        else
-          raise ArgumentError, "Only mysql_native_password (*ABCD...XXX) hashes are supported"
```
hunner: Just eyeballing it, but it looks like this removes validation for accidentally passing in non-hash passwords, which could cause confusion later down the road. Perhaps the validation should still be performed unless `@resource[:plugin] == 'sha256_password'`?

hunner: According to the docs https://dev.mysql.com/doc/refman/5.7/en/create-user.html the forms are `IDENTIFIED WITH auth_plugin BY 'auth_string'` and `IDENTIFIED WITH auth_plugin AS 'hash_string'`

Your pull request only uses the `AS` keyword so I'm guessing is still hashed, but the example you gave in the description is not hashed. Should they be?

Would it make more sense to introduce an `auth_string` parameter for strings other than password hashes?

david22swan:
@jhriggs Apologies but due to the amount of time that has gone by without you responding to the above requested changes we feel that we must close this PR. If in the future you wish to continue on this work we will of course be happy to review it once again.
Best Wishes

PR #934, opened 2017-03-15
(#4534) Add PROXY grant support to mysql_grant

jhriggs:
The mysql_grant type does not currently support PROXY grants. The only real difference between PROXY and other grants is that the "table" is actually a user (proxy_user@proxy_host). This patch adds this support and allows for the following:

```puppet
mysql_grant { 'foo@bar/baz@blah':
  ensure     => present, # or absent
  user       => 'foo@bar',
  table      => 'baz@blah',
  privileges => 'PROXY',
}
```


wilson208:
👍 LGTM. Thanks for the contribution!

PR #943, opened 2017-04-03
(Modules-4320) Readme update for i18n Process

HelenCampbell:
This is prepping the README for translation. Work carried out include
the following:

Values true, false, undef must be wrapped in backticks.
Strings must always be in single quotes.
Sentences must not contain hard returns; they cause the translation
software to create new paragraphs.
For code blocks, use triple backticks (```) and the name of the
language. Without language highlighting, the backticks get stripped out.
Also, language highlighting is nice.
Standardize exceptions to default values (for example, Default value:
'/etc/ntp.conf' (Solaris: '/etc/inet/ntp.conf').)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,9 @@ This module manages both the installation and configuration of MySQL, as well as
 
 To install a server with the default options:
 
-`include '::mysql::server'`. 
+`include '::mysql::server'`.
 
-To customize options, such as the root
-password or `/etc/my.cnf` settings, you must also pass in an override hash:
+To customize options, such as the root password or `/etc/my.cnf` settings, you must also pass in an override hash:
 
 ```puppet
 class { '::mysql::server':
@@ -42,7 +41,7 @@ class { '::mysql::server':
 }
 ```
 
-See [**Customize Server Options**](#customize-server-options) below for examples of the hash structure for $override_options`.
+See [**Customize Server Options**](#customize-server-options) below for examples of the hash structure for $override_options.
 
 ## Usage
 
@@ -67,9 +66,9 @@ For options that you would traditionally represent in this format:
 thing = X
 ```
 
-...you can make an entry like `thing => true`, `thing => value`, or `thing => "` in the hash. Alternatively, you can pass an array, as `thing => ['value', 'value2']`, or list each `thing => value` separately on separate lines. 
+Entries can be created as `thing => true`, `thing => value`, or `thing => ""` in the hash. Alternatively, you can pass an array as `thing => ['value', 'value2']` or list each `thing => value` separately on individual lines.
 
-You can pass a variable in the hash without setting a value for it; the variable would then use MySQL's default settings. To exclude an option from the my.cnf file --- for example, when using `override_options` to revert to a default value --- pass `thing => undef`.
+You can pass a variable in the hash without setting a value for it; the variable would then use MySQL's default settings. To exclude an option from the `my.cnf` file --- for example, when using `override_options` to revert to a default value --- pass `thing => undef`.
 
 If an option needs multiple instances, pass an array. For example,
 
@@ -83,7 +82,7 @@ $override_options = {
 
 produces
 
-```
+```puppet
 [mysqld]
 replicate-do-db = base1
 replicate-do-db = base2
@@ -147,7 +146,7 @@ To add custom MySQL configuration, place additional files into `includedir`. Thi
 
 To instantiate databases and users on an existing MySQL server, you need a `.my.cnf` file in `root`'s home directory. This file must specify the remote server address and credentials. For example:
 
-```
+```puppet
 [client]
 user=root
 host=localhost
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,10 +29,9 @@ This module manages both the installation and configuration of MySQL, as well as
 
 To install a server with the default options:
 
-`include '::mysql::server'`. 
+`include '::mysql::server'`.
 
-To customize options, such as the root
-password or `/etc/my.cnf` settings, you must also pass in an override hash:
+To customize options, such as the root password or `/etc/my.cnf` settings, you must also pass in an override hash:
```
pmcmaw: Single quotes around the '/etc.my.cnf'.
Example from ntp is: '/etc/ntp.conf'

README.md
```diff
@@ -67,7 +66,7 @@ For options that you would traditionally represent in this format:
 thing = X
 ```
 
-...you can make an entry like `thing => true`, `thing => value`, or `thing => "` in the hash. Alternatively, you can pass an array, as `thing => ['value', 'value2']`, or list each `thing => value` separately on separate lines. 
+Entries can be created as `thing => true`, `thing => value`, or `thing => ""` in the hash. Alternatively, you can pass an array as `thing => ['value', 'value2']` or list each `thing => value` separately on individual lines.
 
 You can pass a variable in the hash without setting a value for it; the variable would then use MySQL's default settings. To exclude an option from the my.cnf file --- for example, when using `override_options` to revert to a default value --- pass `thing => undef`.
```
pmcmaw: Add single quotes around 'my.cnf' so that it is not translated.

README.md
```diff
@@ -523,50 +558,71 @@ Directory in which to store backups.
...
+Defaults to true.
 
 ##### `backuprotate`
 
-How many days to keep backups. Valid value is an integer. Defaults to '30'.
+How many days to keep backups.
+
+Valid value is an integer.
+
+Defaults to '30'.
```
jbondpdx: This 30 is a numeric, so actually shouldn't have single quotes. I know you didn't add them, but please make them go away?

README.md
```diff
@@ -523,50 +558,71 @@ Directory in which to store backups.
...
 
 ##### `delete_before_dump`
 
-Whether to delete old .sql files before backing up. Setting to true deletes old files before backing up, while setting to false deletes them after backup. Valid values are true, false. Defaults to false.
+Whether to delete old .sql files before backing up. Setting to true deletes old files before backing up, while setting to false deletes them after backup.
+
+Valid values are true, false.
```
jbondpdx: `true`, `false`.

README.md
```diff
@@ -523,50 +558,71 @@ Directory in which to store backups.
...
+
+Defaults to false.
```
jbondpdx: `false`.

README.md
```diff
@@ -523,50 +558,71 @@ Directory in which to store backups.
...
 ##### `backupdatabases`
 
 Specifies an array of databases to back up.
 
 ##### `file_per_database`
 
-Whether a separate file be used per database. Valid values are true, false. Defaults to false.
+Whether a separate file be used per database.
+
+Valid values are true, false.
```
jbondpdx: `true`, `false`.

README.md
```diff
@@ -523,50 +558,71 @@ Directory in which to store backups.
...
+
+Defaults to false.
```
jbondpdx: `false`.

Actually, I'm going to stop marking these. But anyway, the values true and false should have backticks, except within a code sample

README.md
```diff
@@ -612,19 +668,25 @@ The password to create for MySQL monitoring.
...
 #### mysql::server::mysqltuner
 
 **Note**: If you're using this class on a non-network-connected system, you must download the mysqltuner.pl script and have it hosted somewhere accessible via `http(s)://`, `puppet://`, `ftp://`, or a fully qualified file path.
 
 ##### `ensure`
 
-Ensures that the resource exists. Valid values are `present`, `absent`. Defaults to `present`.
+Ensures that the resource exists.
+
+Valid values are `present`, `absent`.
```
jbondpdx: present and absent are strings, not true Booleans, so they get single quotes

README.md
```diff
@@ -785,50 +936,72 @@ mysql_database { 'mysql':
...
 
 The password for $user for the database you're creating.
 
 ##### `dbname`
 
-The name of the database to create. Defaults to $name.
- 
+The name of the database to create.
+
+Defaults to $name.
```
jbondpdx: OK, I think $name is still a string, but it has a variable, so let's do double quotes around it.

jbondpdx: And change $user in the `password` setting right above this, in the same way.

README.md
```diff
@@ -785,50 +936,72 @@ mysql_database { 'mysql':
...
+Valid values are 'present', 'absent'.
+
+Defaults to 'present'.
 
 ##### `import_timeout`
 
-Timeout, in seconds, for loading the sqlfiles. Defaults to '300'.
+Timeout, in seconds, for loading the sqlfiles.
+
+Defaults to '300'.
```
jbondpdx: 300 here is a numeric, so no single quotes or backticks.

README.md
```diff
@@ -1026,14 +1226,9 @@ Testing on other platforms has been minimal and cannot be guaranteed.
 
 ## Development
 
-Puppet Labs modules on the Puppet Forge are open projects, and community
-contributions are essential for keeping them great. We can't access the
-huge number of platforms and myriad of hardware, software, and deployment
-configurations that Puppet is intended to serve.
+Puppet Labs modules on the Puppet Forge are open projects, and community contributions are essential for keeping them great. We can't access the huge number of platforms and myriad of hardware, software, and deployment configurations that Puppet is intended to serve.
```
jbondpdx: Oops, delete "Labs" from this, please!

HelenCampbell:
@jbondpdx Ready for re-review! Remind me to squash the commits before this gets merged :)

HelenCampbell:
@jbondpdx Done!

PR #945, opened 2017-04-18
(MODULES-4743) mysql : cannot initialize database dir not empty

shawnferry:
(MODULES-4743) puppetlabs-mysql : cannot initialize database dir not empty


pmcmaw:
@shawnferry thank you for your contribution, it is much appreciated! 👍

PR #947, opened 2017-04-25
Add --incremental-basedir option to xtrabackup cron job

jandro:
Incremental backup with xtrabackup is not working unless the option --incremental-basedir is passed on to the cronjob.
It will also take the last created backup within $backupdir as the base for the new incremental backup.

### DIFF
--- a/manifests/backup/xtrabackup.pp
+++ b/manifests/backup/xtrabackup.pp
@@ -57,7 +57,7 @@ class mysql::backup::xtrabackup (
 
   cron { 'xtrabackup-daily':
     ensure  => $ensure,
-    command => "/usr/local/sbin/xtrabackup.sh --incremental ${backupdir}",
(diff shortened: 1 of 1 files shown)


hunner:
What if there has never been a backup? My guess would be that `${backupdir}` would be empty and thus it would execute `--incremental-basedir ${backupdir}/` though there could be situations when the files in the backup dir are not actually backups, or are backups from crufty old systems.

rgevaert:
FYI if there haven't been any backups it reports, it complains:
```
xtrabackup: Error: cannot open /var/backups/mysql//xtrabackup_checkpoints
xtrabackup: error: failed to read metadata from /var/backups/mysql//xtrabackup_checkpoints
```
So in fact you will not have any backups unless the first fill backup is run.  I guess that is OK if it is mentioned in the documentation.

david22swan:
@jandro Is there any movement on this PR?

david22swan:
@jandro I apologize but due to the lack of movement on this PR and your lack of response when prompted we feel that we must know close this PR. If at a later time you return to and complete it then we will of course e more than happy to once again review it.
Best Wishes

PR #959, opened 2017-06-15
versioncmp parameter compatibility

devcfgc:
The error I was getting was:
```
Error: Evaluation Error: Error while evaluating a Function Call, 'versioncmp' parameter 'a' expects a String value, got Undef at '/puppetlabs-mysql/manifest/params.pp'....
```
Puppet 4 introduces strong type checking so after this change it will work on puppet 4 and puppet 3
from
`if versioncmp($::operatingsystemmajrelease, '14.10') > 0 {`
to
`if versioncmp("${::operatingsystemmajrelease}", '14.10') > 0 {`

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -385,7 +385,9 @@ class mysql::params {
 
   case $::operatingsystem {
     'Ubuntu': {
+      # lint:ignore:only_variable_string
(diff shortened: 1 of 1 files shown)


igalic:
@devcfgc thanks a lot for catching & fixing this!

PR #961, opened 2017-06-27
(MODULES-4604) move name validation in mysql_grant type

eputnam:
Users are running into an issue where \`puppet resource mysql_grant\` fails because there is a global name validation using properties that only exist in a compiled catalog and not in the naked type that is instantiated during a \`puppet resource\` run. self[:user] and self[:table] are nil in this use case so there is a failure. This commit surrounds the validation in an if statement so it is not called during \`puppet resource\`. Also, a happy path test is added and the acceptance test is removed."


LichP:
This validation still erroneously fails if `user` is passed in as an uninterpolated variable. For example:

```
$user_host = 'someuser@somehost'
$db_name = 'somedb'
mysql_grant { "${user_host}/${db_name}.*":
  ensure     => present,
  privileges => $tables_or_all,
  table      => "${db_name}.*",
  user       => $user_host,
}
```
The above snippet fails the validation check on 5.2.0 unless the user line is changed to `user => "${user_host}",`

PR #965, opened 2017-07-26
mark Debian/stretch as supported

mmoll:


metadata.json
```diff
@@ -57,6 +57,7 @@
         "6",
         "7",
         "8"
```
ekohl: You forgot a comma that I included in https://github.com/puppetlabs/puppetlabs-mysql/pull/966.

mmoll:
let's take yours then :)

PR #972, opened 2017-08-14
Fixing empty user/password issue

ajardan:
With empty strings the template has the --user and --password options, but those are empty.

### DIFF
--- a/manifests/backup/xtrabackup.pp
+++ b/manifests/backup/xtrabackup.pp
@@ -1,8 +1,8 @@
 # See README.me for usage.
 class mysql::backup::xtrabackup (
   $xtrabackup_package_name = $mysql::params::xtrabackup_package_name,
-  $backupuser              = '',
-  $backuppassword          = '',
+  $backupuser              = undef,
+  $backuppassword          = undef,
   $backupdir               = '',
   $maxallowedpacket        = '1M',
   $backupmethod            = 'mysqldump',


igalic:
```ruby
[2] pry(main)> pw = ''
=> ""
[3] pry(main)> if pw then :true else :false end
=> :true
[4] pry(main)> 
```
indeed. @ajardan how will this impact the other backup methods / templates?
they have the same defaults, and should possibly get the same treatment…

from what i gather, the easiest would perhaps even be to not give it any default at all, since a backupuser / backuppassword is always required. or?

ajardan:
@igalic for xtrabackup it is not required, since this doesn't need a MySQL connection. But it is required for mysqldump, so the other 2 methods are fine.

And yes, I agree, would be better to keep them all undef, and require/fail wherever it is necessary to perform the backup.

igalic:
could you please update your commit message to explain this.

manifests/backup/xtrabackup.pp
```diff
@@ -1,8 +1,8 @@
 # See README.me for usage.
 class mysql::backup::xtrabackup (
   $xtrabackup_package_name = $mysql::params::xtrabackup_package_name,
-  $backupuser              = '',
-  $backuppassword          = '',
+  $backupuser              = undef,
+  $backuppassword          = undef,
```
rnelson0: It seems like this might be a good chance to start using types and label these as `Optional[String]`.

igalic: here it would be okay

but not in the other backup classes.

ajardan:
@igalic I updated my commit message.

@rnelson0 I would be happy to do so, but we are still using Puppet 3.8.x, and I guess data types are part of Puppet 4.x ?

rnelson0:
@ajardan Data types are available in Puppet 3 with the future parser.

PR #975, opened 2017-08-23
(MODULES-5528) mysql_install_db should be optional

jarretlavallee:
Prior to this PR, `mysql_install_db` was a hard requirement in the
mysql provider. Since this command is not in all versions of MySQL, this
commit removes the hard requirement of `mysql_install_db`. Existing
conditional logic already exists to handle situations where the
`mysql_install_db` is not used.


HelenCampbell:
Going to get this out with a release so for speeds sake I made a new PR with your change:
https://github.com/puppetlabs/puppetlabs-mysql/pull/990. I'll go ahead and close this. Thanks!

PR #984, opened 2017-09-06
MODULES-5405 interpolation for puppet strings

tphoney:


tphoney:
We need a release of the translate module.

PR #993, opened 2017-09-13
(MODULES-5618) Hide logging of password_hash changes in mysql::user

jhriggs:


HAIL9000:
Hey @jhriggs, thanks for this pull request!

One thing I noticed is that the rubocop job is failing for a few reasons (you can see the full output of that [here](https://travis-ci.org/puppetlabs/puppetlabs-mysql/jobs/275035854)). If you wouldn't mind addressing those it'd be greatly appreciated as it helps keep our code tidy and consistent. Plus it does look like the methods you added take a variable that they're not using?

Let me know if you have any further questions!

jhriggs:
Updated. For consistency, I had literally copied the code from Puppet's `lib/puppet/type/user.rb`.

jhriggs:
The last remaining "offense" in Rubocop cannot be "fixed".

lib/puppet/type/mysql_user.rb
```diff
@@ -46,6 +46,18 @@
   newproperty(:password_hash) do
     desc 'The password hash of the user. Use mysql_password() for creating such a hash.'
     newvalue(%r{\w*})
+
+    def change_to_s(currentvalue, _newvalue)
+      (currentvalue == :absent) ? 'created password' : 'changed password'
+    end
+
```
yakirgb: Based on Robocop - [PredicateName](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/PredicateName) you cannot use `is_`, you can fix it by disable `Style/PredicateName` and enable it after `is_to_s`
`# rubocop:disable Style/PredicateName`

lib/puppet/type/mysql_user.rb
```diff
@@ -46,6 +46,18 @@
...
+    def is_to_s(_currentvalue)
+      '[old password hash redacted]'
+    end
+
```
yakirgb: # rubocop:enable Style/PredicateName

lib/puppet/type/mysql_user.rb
david22swan: This look's good. As a note for the future though you can put the disable on the line that violates the rule itself. This will make it only disable for that line. It's only if it is placed on an empty line that it need's to be reenabled.

PR #997, opened 2017-09-19
(FM-6362) Install i18n gems from RubyGems

HAIL9000:
Prior to this commit, our i18n gems were being installed from git
as they were not yet released. Now that they've been released,
just install them from RubyGems.

_This PR is queued up in advance but shouldn't be merged until all the gems have been released._

PR #998, opened 2017-09-21
(FM-6362) - Updating i18n gems to point at Rubygems

pmcmaw:
Currently the gems are pointing to git repos, now they have been
released we want them pointing at Rubygems.


pmcmaw:
Closing at this PR duplicates #997

PR #1007, opened 2017-10-05
Updating translations for locales/ja/puppetlabs-mysql.po

HelenCampbell:


pmcmaw:
Merging as this PO file needs to be in master, then I will address the errors in the next PR.

PR #1015, opened 2017-10-09
WIP - Adding a test using beaker-testmode_switcher

pmcmaw:
All current tests are applying the manifest. 
Implementing 'beaker-testmode_switcher' provides us with the capability of injecting the manifest to the master and running 'puppet agent -t' on the agent. This exercises the interface between the master and agent and tests more like an end user.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,13 @@ matrix:
   include:
   - rvm: 2.3.1
     dist: trusty
-    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/ubuntu-14.04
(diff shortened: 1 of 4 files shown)


pmcmaw:
I need to work out a way to run the test if there is a master/agent setup present, pe_install isn't the best method to determine this.

hunner:
@pmcmaw Do you want to get this on the roadmap with Mallon? Did we pass over it during some sprint and need to pick it back up?

pmcmaw:
@hunner Im going to close this for now.
If this comes back up I will be able to use this PR as a start point. 
:-)

PR #1019, opened 2017-11-06
(MODULES-4793) replace characters that are invalid for paths 

tun0:
Replace characters that are invalid for paths.

This patch fixes: [MODULES-4793](https://tickets.puppetlabs.com/browse/MODULES-4793)

### DIFF
--- a/manifests/server/mysqltuner.pp
+++ b/manifests/server/mysqltuner.pp
@@ -7,8 +7,8 @@ class mysql::server::mysqltuner(
 ) {
 
   if $source {
-    $_version = $source
-    $_source  = regsubst($source,'[:/]','_','G')
+    $_version = regsubst($source,'[:/]','_','G')
+    $_source  = $source
   } else {
     $_version = $version
     $_source  = "https://github.com/major/MySQLTuner-perl/raw/${version}/mysqltuner.pl"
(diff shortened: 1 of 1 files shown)


pmcmaw:
Closing and reopening to resolve Travis/bundler issues, this is caused due to https://tickets.puppetlabs.com/browse/MODULES-6339.

manifests/server/mysqltuner.pp
```diff
@@ -8,7 +8,7 @@
 
   if $source {
     $_version = $source
```
hunner: Does that change not need to go on the `$_version` line rather than the `$_source` line? It appears that the version is the part that breaks.

rlaban:
Ugh, you're correct obviously. I recall having some issues with git and creating the pull request back then. Probably applied the change back and forth manually. And at some point this mistake creeped in.

hunner:
@rlaban Thanks! And now it looks like the unit tests will have to be updated to look for the new titles without the /'s in them to get travis to pass.

david22swan:
@tun0 I apologize but due to the lack of movement on this PR, as well as your failure to respond when poked I a afraid that I must now close this PR. If you wish to later on return and complete this work we will of course be happy to once again review your work.
Best Wishes

PR #1022, opened 2017-11-14
MODULES-6018 properly parse/escape whitespace in tls_options

beatchristen:
1. use tab to separate fields coming from the mysql.user table.
2. wrap tls_options like subject, issuer and cypher with "..."

see https://tickets.puppetlabs.com/browse/MODULES-6018

fixes #1021


pmcmaw:
Hey @beatchristen,

If you are able to add a few tests I will get this change merged for you. Thank you for taking the time to contribute. 

:-)

beatchristen:
will do and makes sense. noticed the \n is now producing an issue.

PR #1031, opened 2017-12-06
Do not run task setup on LTS

pmcmaw:
Currently task setup is trying to run on LTS which causes failures as it
is not supported.


pmcmaw:
I have ran this through experimental with no issues and is passing on LTS.

pmcmaw:
@willmeek you happy with this change?

PR #1035, opened 2017-12-07
#puppethack allow undef value for bind-address

JvGinkel:
https://tickets.puppetlabs.com/browse/MODULES-3168

If set `bind-address` to `undef` the bind-address will not be set in the mysql-config file so MySQL is listening on all interfaces.

```puppet
class { '::mysql::server':
  override_options => {
    mysqld => {
      bind-address => undef,
    },
}
```

### DIFF
--- a/spec/classes/mysql_server_spec.rb
+++ b/spec/classes/mysql_server_spec.rb
@@ -97,6 +97,21 @@ describe 'mysql::server' do
 
           it { is_expected.to contain_file('/tmp/error.log') }
         end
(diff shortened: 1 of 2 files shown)


eputnam:
@JvGinkel we'd be more than happy to merge your tests in but it seems like the template change is not necessary. If you run the tests without the change, they still pass.

JvGinkel:
@eputnam you are right! I have revert the template and fix the rubocop issues.

PR #1040, opened 2017-12-18
[FM-6584] Task test cleanup

willmeek:


spec/acceptance/mysql_backup_spec.rb
```diff
@@ -48,7 +48,7 @@ class { 'mysql::server::backup':
     end
 
     it 'dumps all databases to single file #stdout' do
-      if version_is_greater_than('5.7.0')
+      unless version_is_greater_than('5.7.0')
```
eputnam: was this intentional? it's broken jenkins.

willmeek: Yes, this is to *fix* Jenkins - it was introduced in a recent rubocopificiation work, I added the fix to this PR.

You can view an adhoc run of this branch on Jenkins - https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/linux/view/mysql/view/adhoc/ - all green :)

PR #1044, opened 2018-01-15
Use puppet4 functions-api

juliantodt:
The legacy puppet3 functions-api should be avoided.
This migrates the existing functions to the new api and uses it in the module. It also adds the tests for the new functions based on those for the olds when they existed.
We should keep the legacy functions for now to allow compatibility for those using the functions outside out this module.

### DIFF
--- a/examples/mysql_user.pp
+++ b/examples/mysql_user.pp
@@ -6,18 +6,18 @@ class { 'mysql::server':
 
 mysql_user{ 'redmine@localhost':
   ensure        => present,
-  password_hash => mysql_password('redmine'),
+  password_hash => mysql::password('redmine'),
   require       => Class['mysql::server'],
 }
 
 mysql_user{ 'dan@localhost':
   ensure        => present,
-  password_hash => mysql_password('blah')
+  password_hash => mysql::password('blah')
 }
 
 mysql_user{ 'dan@%':
   ensure        => present,
-  password_hash => mysql_password('blah'),
+  password_hash => mysql::password('blah'),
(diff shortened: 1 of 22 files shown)


juliantodt:
Fixed the rubocop issues that occurred, but didn't know how to replace the `subject.execute` calls in the tests with them still working. Please let me know if you have a solution for this.

david22swan:
@juliantodt In regards to the rubocop issue for now please add it to the rubocop_todo file, this will allow your test's to pass. Please note that this should be a last option though as the aim is for the todo file to be blank whenever possible.

david22swan:
@juliantodt Are you still active on this pr?
In regards to the rubocop issue at this point I feel that you should disable the rule for now, an example on how to do so is shown below:

Disable Commands:
Disable rubocop over a block off code
   # rubocop:disable Lint/UselessAssignment
        {Code Block}
    # rubocop:enable Lint/UselessAssignment

Disable rubocop for a single line:
    {Code Line} # rubocop:disable RSpec/InstanceVariable

juliantodt:
Fixed the rubocop issues like @david22swan suggested and rebased the pr.

juliantodt:
@hunner Added the deprecation warnings. Couldn't use stdlib's deprecation function because we allow old enough versions of it together with the mysql-module though.

eputnam:
@juliantodt went ahead and added strings docs to have them in there. It's a new pseudo-requirement so I didn't think you needed to be responsible for it.

PR #1047, opened 2018-01-23
(maint) Plugins don't exist on 5.1; password field name changed

hunner:


tasks/export.rb
```diff
@@ -4,13 +4,13 @@
...
 def get(file, database, user, password)
-  cmd_string = "mysqldump"
+  cmd_string = 'mysqldump'
   cmd_string << " --databases #{database}" unless database.nil?
   cmd_string << " --user=#{user}" unless user.nil?
   cmd_string << " --password=#{password}" unless password.nil?
   cmd_string << " > #{file}" unless file.nil?
-  stdout, stderr, status = Open3.capture3(cmd_string)
-  raise Puppet::Error, _("stderr: '#{stderr}'") if status != 0
+  stdout, _stderr, status = Open3.capture3(cmd_string)
```
alexjfisher: @hunner This wasn't a hurtful cop.  There's a misplaced `"` on the line below.  That's why rubocop was complaining about `stderr` not being used.

PR #1048, opened 2018-01-25
(MODULES-6454) - PDK Convert mysql

pmcmaw:
Using PDK to convert module so that it is compliant with the PDK.

The second commit addresses a bundler issue MODULES-6339.
The next time PDK convert is ran it will re-add the newer version of the bundler gem.


.rubocop.yml
```diff
@@ -1,8 +1,7 @@
 ---
-require:
- - rubocop-rspec
- - rubocop-i18n
```
pmcmaw: @HelenCampbell we may need this.

pmcmaw:
`GetText/DecorateString` is disabled as this rule flags that `desc` and `@doc`. Currently this is not something we need to decorate. 

I have tested this PR to ensure failures are caught in Travis when necessary ruby and puppet strings are not decorated.

Ruby:
```
bundle exec rake rubocop
lib/puppet/provider/mysql.rb:63:28: C: GetText/DecorateFunctionMessage: 'raise' function, message should be decorated. 
      raise Puppet::Error, "#mysql_caller: Unrecognised type '%{type}'" % { type: type }
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Puppet:
```
bundle exec rake release_checks
manifests/server.pp - WARNING: 'warning' messages should be decorated: eg translate(The `old_root_password` attribute is no longer used and will be removed in a future release.) on line 48
```

.sync.yml
```diff
@@ -1,13 +1,69 @@
...
+  spec_overrides:
+    -  "require 'spec_helper_local'"
+
 .rubocop.yml:
-  unmanaged: true
-.rubocop_todo.yml:
-  unmanaged: true
+   default_configs:
+    inherit_from: .rubocop_todo.yml
+    require:
```
pmcmaw: By adding the `rubocop-rspec` require here it allows us to require multiple things. If I only require `rubocop-i18n` the `rubocop-rspec` from the template it overwritten.

pmcmaw: `rubocop-i18n` is required because this module has been internationalized.

.sync.yml
```diff
@@ -1,13 +1,69 @@
...
+spec/spec_helper.rb:
+  spec_overrides:
+    -  "require 'spec_helper_local'"
+
 .rubocop.yml:
-  unmanaged: true
-.rubocop_todo.yml:
-  unmanaged: true
+   default_configs:
+    inherit_from: .rubocop_todo.yml
```
pmcmaw: The `.rubocop_todo.yml` is required as it contains configuration to disable a cop. The cop is currently not needed.

.sync.yml
```diff
@@ -1,13 +1,69 @@
...
+        from_env: BEAKER_RSPEC_VERSION
+    ':development':
+      - gem: puppet-blacksmith
+        version: '~> 3.4'
+      - gem: puppet-lint-i18n
+
+Rakefile:
+  requires:
+    - puppet_blacksmith/rake_tasks
+    - puppet_pot_generator/rake_tasks
```
pmcmaw: This is required for i18n.

pmcmaw:
@HelenCampbell I have updated this PR with all the necessary changes to ensure that it has all the i18n tools and that they work. I have commented on changes that differ compared to the usual PDK convert PRs. It would be great if I could get another review now this has all been done.

pmcmaw:
This has ran through the adhoc pipeline with no issues.

Rakefile
```diff
@@ -1,37 +1,4 @@
 require 'puppetlabs_spec_helper/rake_tasks'
-require 'puppet-lint/tasks/puppet-lint'
```
HelenCampbell: This needs added to .sync.yml

PR #1049, opened 2018-02-06
Compatibility for Alpine linux

cisco87:


pmcmaw:
Hey @cisco87,
When merging this PR I would like to make you aware this will make the code base compatible with the module, however Puppet will not support this operating system. By support I mean we will not be running tests against this OS or providing bug fixes specific to this OS. I would like to make sure you are ok with this before I merge. 👍

eputnam:
@cisco87 please see @pmcmaw 's comment. also, if you'd like to add an alpine docker nodeset in spec/acceptance/nodesets, feel free to do so!

cisco87:
Hello, sorry for the delay on the reply, I've been away from github a bit. I will try to maintain it updated with changes in both alpine/puppet-mysql, I use it every day so I should be quite able to catch any regression. I will have a look at acceptance testing approach, thanks for the merging

PR #1057, opened 2018-03-13
fix archlinux compatibility

bastelfreak:
I added the variables daemon_dev_package_name, client_dev_package_name and includedir for Archlinux in the params.pp file. I also disabled Archlinux for tests on two classes, they don't support archlinux.


bastelfreak:
oh ffffff rubocop.....¯\_(ツ)_/¯

PR #1058, opened 2018-03-20
documentation: provide a more useful Hiera example

joekohlsdorf:
The previous example for configuring the module with Hiera wasn't a pure Hiera example but required additional custom module code. It also overwrote default settings for a specific OS and use-case which isn't necessary when using the module.

My example shows the use-case which 95% of users have:
* Configure the minimal server settings necessary to run.
* Set up a user, a DB and a grant on it for the user.

Note: The password of the password hash in the example is `password`, a hash has to be used because the provider doesn't allow a plaintext password. For me this is acceptable but unfortunately inconsistent with the configuration of the `root_password`.
Sadly I don't have time to implement a real solution: Making `myql::db` configurable from Hiera.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,66 +320,29 @@ Class['apt::update'] ->
 Class['::mysql::client']
 ```
 
-### Install MySQL Community server on CentOS
+### Using the module with Hiera
 
-You can install MySQL Community Server on CentOS using the mysql module and Hiera. This example was tested with the following versions:
-
-* MySQL Community Server 5.6
-* Centos 7.3
-* Puppet 3.8.7 using Hiera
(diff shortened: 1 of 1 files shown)


pmcmaw:
Hey @joekohlsdorf apologies for the delay. 

Would it be possible to rebase your example and I will work on getting it merged.

Also Author names with special characters seem to have been updated with question marks. If you could resolve this it would be great 👍

david22swan:
@joekohlsdorf Apologies but are you still intending to finish this pr??

joekohlsdorf:
I am not interested in supporting this project anymore, thanks.

PR #1063, opened 2018-04-09
(MODULES-6981) Do not try to read ~root/.my.cnf when calling "mysqld -V"

simondeziel:


pmcmaw:
Apologies for the slow feedback, I have closed and reopened to kick off tests on a more up to date test matrix. However I have seen that this change causes a test failure. Before being able to merge would it be possible to address the error?

simondeziel:
All tests pass now, thanks.

PR #1064, opened 2018-04-23
(MODULES-6627) Removes unused --host flag from mysqlcaller

HelenCampbell:


hunner:
Odd since it doesn't appear to be used for anything. It was added as part of b3658f8 with this comment:
> Now the options to all mysql calls can be modified from one place.  This will be useful when we are converting the module to allow for databases on multiple remote servers, since we will be able to add the -h switch to all calls at once.

Perhaps it snuck in as an unfinished feature that the user was working on? Anyway, it can be added "in one place" as the comment says later if needed.

PR #1086, opened 2018-05-30
[FM-6962] Removal of unsupported OS from mysql

david22swan:


david22swan:
https://github.com/puppetlabs/ci-job-configs/pull/4562

PR #1087, opened 2018-06-07
Create users if not exists : Update mysql.rb 

cesarpball:
We are having issues when we have a master-slave replication and puppet runs first on the slave instance. After running in the slave, and then in the master, the replica is broken because the user is already created. So if the user already exists, in my opinion, we shouldn't do anything.

Any thoughts?


david22swan:
@cesarpball The test's run against your pr are currently failing

david22swan:
@cesarpball Is there any movement on this PR?

pmcmaw:
Closing as per the above comment.
I would like to reiterate if you believe this is the incorrect decision please feel free to reopen. 

We also would like to thank you for taking the time to contribute to this module.

Many Thanks
The Modules Team.

PR #1089, opened 2018-06-17
Lookup override_options as a hiera_hash so it can be defined across multiple levels of hiera

cdloh:
Presently the override options can only be set at one level of hiera. The top level of hiera wins when it does the lookup.

This means that you cannot define it across multiple levels of hiera at all which makes it a pain to manage clusters of MySQL servers.

E.G. usage case: one master server and one slave server in a cluster.

Within a cluster hiera file you can set all of the tuning and server mysql settings to include in my.cnf eg

```
    query_cache_limit: 128M
    query_cache_size:  640M
    join_buffer_size: 64M
    table_cache: 500
```

However since the bind-address & server-id need to be different between the two servers you cannot set those two parameters within the same hiera file.

Moving them into node specific hiera files then makes this possible. However presently the module would then only pickup the server hiera settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,14 @@ class { '::mysql::server':
 }
 ```
 
+Or add overrides within your hiera by addin `mysql::server::override_options` eg
+
+```yaml
+mysql::server::override_options:
+  mysqld:
+    pid-file:  '/tmp/mysql.pid'
+```
+
 See [**Customize Server Options**](#customize-server-options) below for examples of the hash structure for $override_options.
 
 ## Usage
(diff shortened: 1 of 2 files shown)


david22swan:
@cdloh Please update the docs to match your change and add some test coverage.

cdloh:
@david22swan I've updated the documentation. Let me know if any of it isn't clear.

What kind've tests are you looking for? Should I load some sample hiera during the tests?

david22swan:
@cdloh You should take a look at the existing server_spec.rb tests:
https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/classes/mysql_server_spec.rb
https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/acceptance/mysql_server_spec.rb
These show you how the test's should look.

david22swan:
@cdloh Sorry to bother, but I was wondering whether yo were still working on this PR?

pmcmaw:
Hey @cdloh 

Sadly as this PR has sat still for a prolonged period of time I have decided to close this PR. If at any stage you revisit to add tests, please feel free to reopen and we can work with you on getting your piece of work merged.

I would like to thank you for taking the time to contribute to our module. 👍

Many Thanks
The Modules Team

PR #1092, opened 2018-06-22
MySQL 8 compatibility in user management

zpetr:
MySQL 8 doesn't support "... IDENTIFIED BY PASSWORD ..." syntax  and resources options (used after WITH in GRANT command) were moved from GRANT to ALTER USER.
If plugin is ommited, 'mysql_native_password' plugin is used by default (because the default MySQL 8 user plugin is 'caching_sha2_password')


david22swan:
@zpetr Your PR has failed the syntax check.

PR #1103, opened 2018-07-30
Fix mysql_plugin examples and add section in README

martialblog:
Hi, 

we ran into some issues when using the mysql_plugin Type and the documentation wasn't up2date. So we fixed it. 

Cheers,
Markus


HelenCampbell:
The travis failure on this PR is due to it needing some changes that are in master but not in this PR. As a result I'm going to go ahead and rebase&merge since this only readme and example updates.
Thank you for your work @martialblog !

martialblog:
@HelenCampbell Hooray :shipit: Thanks

PR #1107, opened 2018-08-03
Removing query_cache ops that are no longer supported in MySQL >= 8.0

ernstae:
While using MySQL 8.0, the configuration options in the mysqld params section had `query_cache_limit` and `query_cache_size` variables which are no longer supported for MySQL >= 8.0

Moving these into the version-specific hashes, rather than the `mysqld` default hash.  I didn't love having to duplicate those entries, but it does follow the same pattern of the `myisam-recover` option. 🤷‍♂️ 

This results in MySQL 8.0.12 actually starting up, rather than throwing the following error example:

```
[ERROR] [MY-011071] [Server] unknown variable 'query_cache_limit=1M'
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -455,12 +455,18 @@ class mysql::params {
     },
     'mysqld-5.5'       => {
       'myisam-recover' => 'BACKUP',
+      'query_cache_limit'     => '1M',
+      'query_cache_size'      => '16M',
     },
     'mysqld-5.6'              => {
       'myisam-recover-options' => 'BACKUP',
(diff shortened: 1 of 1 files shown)


lroehrs:
To override `query_cache_size` and `query_cache_limit`, you have to set the specific version: 

```
mysql::server::override_options:
  mysqld-5.7:
    query_cache_size: ~
```

The options in `mysqld` will be overrided from the default values: 

https://github.com/puppetlabs/puppetlabs-mysql/blob/main/manifests/params.pp#L491

PR #1108, opened 2018-08-05
This supports the MySQL 8.0 password and user management changes.

ernstae:
This looks like it's actually reasonably similar to #1092 but I worked on this in another environment, and was making an attempt to contribute back upstream after success in my own environment.

The primary change here, is to avoid the non-existent property hash for per-user resource limitations in versions higher than 8.0, and also to handle passwords using the new syntax.


ernstae:
Looks like I'll be working on backwards compatibility, next :wink:

ernstae:
I'm going to work through more testing and possibly update the testing suite to include MySQL 8.0, and see how far I can take this work.

PR #1115, opened 2018-08-29
(MODULES-6923) remove staging module

tphoney:


manifests/server/mysqltuner.pp
```diff
@@ -17,6 +17,7 @@
   $environment = undef, # environment for staging::file
```
ekohl: This parameter is now unused

spec/classes/mysql_server_mysqltuner_spec.rb
```diff
@@ -1,17 +1,12 @@
 require 'spec_helper'
 
 describe 'mysql::server::mysqltuner' do
-  on_supported_os.each do |os, facts|
+  on_supported_os.each do |os, _facts|
```
ekohl: If no fact are set, does it actually matter to test it on every OS? You're effectively testing the same thing multiple times.

tphoney:
Many thanks @ekohl , changes incoming

manifests/server/mysqltuner.pp
```diff
@@ -7,58 +7,31 @@
...
-#   Environment variables active during download, e.g. to download via proxies: environment => 'https_proxy=http://proxy.example.com:80'
-#
 class mysql::server::mysqltuner(
   $ensure  = 'present',
   $version = 'v1.3.0',
   $source  = undef,
-  $environment = undef, # environment for staging::file
 ) {
-
+  $tuner_location = '/usr/local/bin/mysqltuner'
```
ekohl: I tend to prefer something like this as a class parameter. This gives some more flexibility to users

manifests/server/mysqltuner.pp
```diff
@@ -7,58 +7,31 @@
...
   if $source {
     $_version = $source
```
ekohl: This variable appears to be unused now and can be removed.

manifests/server/mysqltuner.pp
```diff
@@ -7,58 +7,31 @@
...
     $_source  = $source
   } else {
     $_version = $version
     $_source  = "https://github.com/major/MySQLTuner-perl/raw/${version}/mysqltuner.pl"
```
ekohl: Unrelated, but GH also has a raw.github.com that's more efficient for them. Might be worth looking into but doesn't have to be part of this PR.

manifests/server/mysqltuner.pp
```diff
@@ -7,58 +7,31 @@
...
   }
-
   if $ensure == 'present' {
```
ekohl: Given the archive now also has an `ensure`, does it make sense to drop the if and simply have 2 resources which accept `$ensure`?

tphoney:
@ekohl lovely idea, testing it now.

tphoney:
@ekohl hopefully closer ?

ekohl:
Looks good! Bonus points for data types on parameters but yay for the big simplification already :)

metadata.json
```diff
@@ -8,17 +8,13 @@
...
-      "version_requirement": ">= 1.0.0 < 2.0.0"
-    },
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">= 3.2.0 < 5.0.0"
     },
     {
-      "name": "puppet/staging",
-      "version_requirement": ">= 1.0.1 < 4.0.0"
+      "name": "puppetlabs/translate",
```
bastelfreak: the diff would be a tiny bit cleaner if this would still be on top of stdlib, but it doesn't really matter.

manifests/server/mysqltuner.pp
```diff
@@ -7,59 +7,22 @@
...
-      require => Staging::File["mysqltuner-${_version}"],
-    }
-  } else {
-    file { '/usr/local/bin/mysqltuner':
-      ensure => $ensure,
-    }
+  file { $tuner_location:
+    ensure => $ensure,
+    mode   => '0550',
+    source => $_source,
```
bastelfreak: Nice. I expected a really complicated rewrite by replacing staging with archive, but I also forgot that the file resource can now take https links <3. Since which puppet version is this supported? Does the support puppet version range within this module needs to be adjusted?

ekohl:
It'd be nice if this could get merged.

tphoney:
Hi @ekohl and @alexjfisher i changed the commit and pr message. This will get merged once we have got the puppet 6 release for this module done. It will be a major version bump ie X.y.z because of it being backwards incompatible.

manifests/server/mysqltuner.pp
HelenCampbell: Double checked this before merging, but http support for file source was in 4.7 which is the lower boundary of the module: https://puppet.com/docs/puppet/4.7/type.html#file-attribute-source , so we should be good without bumping the puppet version.

PR #1117, opened 2018-09-07
add mysql_native_password for mariadb 10.2 in password_hash

mlk-89:
Fix idempotent problem when mysql user password is changed in mariadb 10.2


gguillotte:
I think #1135 (merged yesterday) addressed this by checking the MariaDB version when determining how to get the password for comparison, while this addresses it by checking the version when updating the password.

PR #1130, opened 2018-09-26
(MODULES-7857) Support user creation on galera

MaxFedotov:
Support user creation on galera. Add IF NOT EXISTS to CREATE USER in mysql_user provider
Description of issue in jira - https://tickets.puppetlabs.com/browse/MODULES-7857

### DIFF
--- a/lib/puppet/provider/mysql_user/mysql.rb
+++ b/lib/puppet/provider/mysql_user/mysql.rb
@@ -69,8 +69,8 @@ Puppet::Type.type(:mysql_user).provide(:mysql, parent: Puppet::Provider::Mysql) 
       end
       @property_hash[:ensure] = :present
       @property_hash[:plugin] = plugin
-    elsif newer_than('mysql' => '5.7.6', 'percona' => '5.7.6')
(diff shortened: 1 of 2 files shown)


david22swan:
@MaxFedotov Could you add some test coverage for your new feature and update the docs please?

MaxFedotov:
@david22swan Hi. Can you please clarify, which docs should be updated? This PR doesn't add new functionality, it's just reuses the same logic, which is already implemented for DROP USER (https://github.com/puppetlabs/puppetlabs-mysql/blob/9ede9a8c3a49ee4ba05ae87ee57073190b8f631a/lib/puppet/provider/mysql_user/mysql.rb#L109). I've looked throught all docs, and wasn't able to find anything related to this.

MaxFedotov:
@david22swan I've rebased on current master and slightly modified my PR and make a few fixes for the last pull request you've merged. There were some places, which had different ALTER commands (IF EXISTS was missing) on the same IF clause, and in some places mariadb was missing in IF clause.
Also i've updated tests.

PR #1136, opened 2018-10-24
Update db.pp

covidium:
Do not send tls_options if the parameter has an undefined value

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -82,7 +82,7 @@ define mysql::db (
       password_hash => mysql::password($password),
     }
   }
-  
+
   ensure_resource('mysql_user', "${user}@${host}", $user_resource)
 
   if $ensure == 'present' {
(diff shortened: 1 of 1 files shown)


covidium:
@igalic unfortunately my Ruby foo is pretty limited and not sure how exactly to do that.

david22swan:
@covidium Your change is currently failing against the travis unit test's, however the error I am seeing is one that I believe was already fixed. As such could you rebase your changes and see if that resolves it. Also it would be good if you could change the name of your commit to one that is more descriptive.

david22swan:
@covidium Is there any movement on this PR?

david22swan:
@covidium Apologies but due to the length of time that this PR has lain fallow, and your lack of response when pinged I feel that I must now close it.
I apologize if this inconveniences you and should you ever decided to return to it we will of course be more than happy to review it again.
Best Wishes

PR #1145, opened 2018-12-07
(MODULES-8193) - Removal of inbuilt deepmerge and dirname functions

david22swan:
Both functions have been removed from the code and their function calls within the manifest's updated as they already exist within puppet itself, in the case of deepmerge and within stdlib in the case of dirname.


manifests/server/config.pp
```diff
@@ -43,9 +43,10 @@
 
     # on mariadb systems, $includedir is not defined, but /etc/my.cnf.d has
     # to be managed to place the server.cnf there
-    $configparentdir = mysql::dirname($mysql::server::config_file)
+    $configparentdir = dirname($mysql::server::config_file)
     if $configparentdir != '/' and $configparentdir != '/etc' and $configparentdir
-        != $includedir and $configparentdir != mysql::dirname($includedir) {
+        != $includedir and $includedir != undef and $includedir != '' and $configparentdir
+        != dirname($includedir) {
```
tphoney: this is hideously hard to read. can it be simplified or split up

david22swan:
Have separated/updated code in such as way as to improve readability while maintaining the original functionality of the code.

david22swan:
@tphoney Are you fine with the code as it is now?

david22swan:
screen shot 2018-12-13 at 12 04 26 pm

Changes run green on adhoc

PR #1147, opened 2018-12-17
Add includedir for Gentoo

baurmatt:
This breaks otherwise for setups using STRICT_VARIABLES=yes


pmcmaw:
Thanks @baurmatt for this PR 👍

PR #1151, opened 2019-01-03
(MODULES-8193) - Wrapper methods created for inbuilt 4.x functions

david22swan:
These were created to bridge the gap between the 4.x and 3.x functions and account for the difference in their naming.


david22swan:
Tested wrappers by replacing mentions of the actual functions within the code.

david22swan:
This pr has been created to allow users to continue to utilize the same function calls after having moved to puppet 4, as the change in naming between the 3.x and 4.x functions would cause breakages i anyone that makes the switch.

tphoney:
Unit tests would be great for this, to show that we at least have the happy path tested.

tphoney:
just waiting on the tests ....

david22swan:
Having some trouble getting the deprecation warning check to work. I've done the wrappers a little differently than how it is in stdlib

tphoney:
So these new wrapper functions, are they mentioned in the readme, do they come out when strings is run ?

david22swan:
Made a strings run, the functions are now in the reference.md.

tphoney:
@david22swan sorry to point it out

PR #1157, opened 2019-01-18
(MODULES-8503) - 7.1.0 Release Prep

david22swan:


pmcmaw:
I don't think doing 2 major releases in a row is a great idea. A chat is required with some tech leads or someone else to get an opinion on this. Possibly a revert of the backwards incompatible change to get feature and bug fix releases out in between. Please sync up with someone else to get their opinion on this.

david22swan:
As a note, we've also got another major release coming once we rip out puppet 4

PR #1160, opened 2019-02-07
fix install ordering for innodb data size

fe80:
Hi,

Since release 5.4.0 MySQL package was install with default configuration, and after the configuration was set.

So we can not change innodb data size configuration for example (the default size is 12Mo and if you change this size MySQL can't be restart).

Someone have notify this conportement on the commit: https://github.com/puppetlabs/puppetlabs-mysql/commit/d13c06b9561d2c216bb9c5b48913ff2b4ac0b41b

I think it's better to rollback the ordering, and eventually change the resource `/etc/mysql/conf.d/mysql.cnf` in post-install class if it's needed

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -143,8 +143,8 @@ class mysql::server (
   }
 
   Anchor['mysql::server::start']
-  -> Class['mysql::server::install']
   -> Class['mysql::server::config']
+  -> Class['mysql::server::install']
   -> Class['mysql::server::binarylog']
   -> Class['mysql::server::installdb']
   -> Class['mysql::server::service']


elfranne:
Any progress on this ?

david22swan:
@fe80 Thank you for your work and we are sorry that the module was broken for you for so long.

yotles:
Hello. Could you please describe why after this changes installation of mysql by this module with version 8.1.0 started to fail? datadir redeclared by $datadir not created before starting of configuring mysql package.

fe80:
Hello @yotles,

We create the datadir before installation in a custom module to create a dedicated logical volume. I think this is why we have this issue.

PR #1161, opened 2019-02-19
[MODULES-8779] Set proper python_package_name for RHEL/CentOS 8

javierpena:
mysql::bindings::python installs the package set by the
python_package_name variable, which is 'MySQL-python' by default
on all RHEL and Fedora-based systems. This is no longer valid for
RHEL 8, since there is no Python3 support for the package.

In RHEL 8, we should use python3-PyMySQL instead.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -49,12 +49,18 @@ class mysql::params {
           } else {
             $provider = 'mysql'
           }
+          $python_package_name = 'MySQL-python'
         }
         /^(RedHat|CentOS|Scientific|OracleLinux)$/: {
           if versioncmp($::operatingsystemmajrelease, '7') >= 0 {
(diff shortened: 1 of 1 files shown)


eimlav:
Hi @javierpena . Thanks for your PR! We do not currently support RedHat 8 however I have tested this change out and it appears to work as intended. Would you be able to add a ticket to your commit message? See [here](https://github.com/puppetlabs/puppetlabs-mysql/blob/master/CONTRIBUTING.md) for more info. Cheers!

javierpena:
Thanks for the tip @eimlav. I have opened the ticket and changed the commit message to include it.

eimlav:
@javierpena Looks good thanks for adding these changes! :)

PR #1163, opened 2019-03-01
Set correct packagename for ruby_mysql on Ubuntu 18.04

datty:
Fails to install ruby_mysql on Ubuntu 18.04, attempts to use the default package name for Debian which isn't correct.


mrunkel:
Any chance this will get merged soon?  I'm having the same problem.

david22swan:
Screen Shot 2019-04-02 at 2 17 44 PM
@datty Your changes has run through clean and looks good to me. Thanks for putting in the work.

PR #1176, opened 2019-03-29
Rotate option for xtrabackup script

elfranne:
Update to xtrabackup:

- set -o pipefail: failed piped command will give a success (pipe worked fine).
- copied the cleanup bits from from the mysqlbackup.sh.erb and the `mkdir -p $DIR`

### DIFF
--- a/templates/xtrabackup.sh.erb
+++ b/templates/xtrabackup.sh.erb
@@ -5,12 +5,37 @@
 <%- end -%>
 #
 # A wrapper for Xtrabackup
-#
+
+ROTATE=<%= [ Integer(@backuprotate) - 1, 0 ].max %>
+DIR=<%= @backupdir %>
+
+# Ensure backup directory exist.
+mkdir -p $DIR
+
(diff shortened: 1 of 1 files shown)


david22swan:
@elfranne This looks like a good feature to me, however in order for it to be merged we would first need default values for the new variables added to the manifest, along with documentation explaining their purpose and how they should be used.
Once this is done this should be able to be merged.

elfranne:
Those option are already in the manifest and are documented:
https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/server/backup.pp
and the provider:
https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/backup/xtrabackup.pp

But they were never actually implemented in the xtrabackup provider, only in mysqldump provider.

david22swan:
Screen Shot 2019-04-01 at 12 37 47 PM
Ok, your changes look good overall and they have passed through the adhoc pipeline cleanly so I'm gonna give this a merge.
Thanks for putting in the work. :)

PR #1181, opened 2019-04-09
(MODULES-8886) Revert removal of deepmerge function

eimlav:
Reverting this change as the deepmerge function in stdlib does not work the same as the mysql deepmerge. As a result, the mysql config is generated with incorrect duplicate entries with '-' and '_' mixed up.

Original PR #1145 @david22swan


eimlav:
Adhoc ran successfully:
<summary><details>Results</details>
Screen Shot 2019-04-12 at 13 24 12
</summary>

PR #1188, opened 2019-04-16
Make incremental backups deactivable

SaschaDoering:
Add a flag for activating or deactivating incremental backups. The
default is true so we don't change anything if you don't use the flag.
If the flag isi false we create a full backup every day instead of one
full backup and six incremental backups a week.

In addition, I outsourced the test of the mysql::backup::xtrabackup
class to test the additional parameters.


carabasdaniel:
👍 LGTM

PR #1189, opened 2019-04-26
Allow own Xtrabackup script

SaschaDoering:
We have our own backup script with some additional features, of which
quite a few are specific for our platform. I could add some of them
here as well, but this would end up in a completely new script.
That's why I try it with this minimally invasive variant :)


tphoney:
great work, thanks for the pr @SaschaDoering !

PR #1190, opened 2019-04-30
Fix the contribution guide URL

mauricemeyer:
While checking the README to verify my next PR is according to the contribution guidelines, I noticed that the link to the contribution guidelines is broken.

This PR rectifies that issue.


tphoney:
thanks for the fix @mauricemeyer ! nice catch !

PR #1192, opened 2019-05-02
Enable module to not use default options

mauricemeyer:
This PR enables you to either use:

* `override_options` as it was the case previously. Those options are merged with the default options
* `options`. Those options won’t be merged with the default options.

Both are mutually exclusive in usage, meaning that one of them has to be empty.

I wrote this PR because we want to get rid of the version specific configuration options as it makes our workflows for changes of default parameters in our profiles easier.

I wrote it like this so that it has clean semantics: 

* override_options tells you that defaults are overridden
* options tells you that options will be set

Therefore, this PR is fully backwards compatible.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -19,6 +19,8 @@
 #   Path to secret file containing temporary root password.
 # @param manage_config_file
 #   Whether the MySQL configuration file should be managed. Valid values are `true`, `false`. Defaults to `true`.
+# @param options
+#   A hash of options structured like the override_options, but not merged with the default options. Use this if you don’t want your options merged with the default options.
 # @param override_options
 #   Specifies override options to pass into MySQL. Structured like a hash in the my.cnf file:  See  above for usage details.
 # @param package_ensure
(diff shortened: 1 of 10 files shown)


manifests/server.pp
```diff
@@ -70,6 +72,7 @@
   $install_options         = undef,
   $install_secret_file     = $mysql::params::install_secret_file,
   $manage_config_file      = $mysql::params::manage_config_file,
+  $options                 = {},
```
bastelfreak: It would be nice if you could add a datatype here that enforces that this is a hash. Also you can enforce the types within the hash

sheenaajay:
Thanks @mauricemeyer for submitting the PR. Could you please incorporate comments. Thank you @bastelfreak for the valuable comments.

manifests/server.pp
mauricemeyer: I added the `mysql::options` data type, is that what you were thinking about?

mauricemeyer:
@sheenaajay Thanks for the reminder! I updated the PR, let’s see what @bastelfreak thinks :)

sheenaajay:
Thank you @mauricemeyer .

manifests/server.pp
bastelfreak: Where is the datatype? I'm unable to find it in your branch.

mauricemeyer: @bastelfreak Sorry, I forgot to push the changes. They are pushed now.

mauricemeyer:
The [test for Puppet 5](https://travis-ci.org/puppetlabs/puppetlabs-mysql/jobs/598599777) failed, but it looks to me like a test problem?

I can’t restart the test, though.

bastelfreak:
@mauricemeyer magic travis trick: close + reopen the PR will retrigger travis.

mauricemeyer:
@bastelfreak Works like a charm, thanks.

lionce:
close and reopen to rerun tests

lionce:
Hello @mauricemeyer ,

It looks good to me! Thank you!

Cheers!

Screenshot 2019-11-07 at 14 06 09

PR #1195, opened 2019-05-14
(MODULES-6875,MODULES-7487) - Fix mariadb mysql_user password idempotency

alexjfisher:
The change introduced in mariadb 10.2.16 (and 10.3.8) where:

> the `SET PASSWORD` command stores user passwords
in the `mysql_native_password` plugin `authentication_string` instead of
in the `password` column

was addressed in https://github.com/puppetlabs/puppetlabs-mysql/pull/1135

This behaviour was short-lived.  In 10.2.19 and 10.3.11 the old
behaviour was restored.

> The fix in the commit dd6e74c62a2 made sure that the password field is always set, because that was the original issue that broke third-party tools.

See https://jira.mariadb.org/browse/MDEV-16774

This commit is based on
https://tickets.puppetlabs.com/browse/MODULES-6875?focusedCommentId=645260&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-645260

### DIFF
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -81,6 +81,14 @@ class Puppet::Provider::Mysql < Puppet::Provider
     self.class.newer_than(forks_versions)
   end
 
+  def self.older_than(forks_versions)
+    forks_versions.keys.include?(mysqld_type) && Puppet::Util::Package.versioncmp(mysqld_version, forks_versions[mysqld_type]) < 0
+  end
+
+  def older_than(forks_versions)
+    self.class.older_than(forks_versions)
+  end
+
   def defaults_file
     self.class.defaults_file
   end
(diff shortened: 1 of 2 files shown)


eimlav:
Hi @alexjfisher . I have come across this same issue while adding RedHat 8 support for the module. 

I ran your fix through adhoc and it worked fine however when I ran against RedHat 8 (which is using MariaDB 10.3.10) the idempotency issues still occured. I added the COALESCE function suggested in MODULES-7487 and this appears to have fixed the failing tests.

Would it therefore be possible to add the COALESCE function at https://github.com/puppetlabs/puppetlabs-mysql/pull/1195/files#diff-c6c504610358af96dcf6e3372321e142R21 ? Cheers!

bodik:
> Would it therefore be possible to add the COALESCE function at https://github.com/puppetlabs/puppetlabs-mysql/pull/1195/files#diff-c6c504610358af96dcf6e3372321e142R21 ? Cheers!

there should be either solution based on version pinning OR the COALESCE, but not the both to keep code somewhat clear

alexjfisher:
@eimlav Is this with RHEL-8 beta?  I'm wondering if they backported the fix into the 10.3.10 rpm??

The final release of RHEL 8 actually ships with 10.3.11, so maybe this isn't a problem after all?

eimlav:
@alexjfisher Ah I see. I'll see if I can try it out on a released version of RHEL8. Thanks for the info!

eimlav:
@alexjfisher Tested out your changes on GA RHEL8 and they passed. Nice work!

PR #1200, opened 2019-05-30
Add option to specify $backupdir as a symlink target, for use with dm…

comport3:
…_crypt encrypted partitions.

### DIFF
--- a/manifests/backup/mysqldump.pp
+++ b/manifests/backup/mysqldump.pp
@@ -72,7 +72,7 @@ class mysql::backup::mysqldump (
     content => template('mysql/mysqlbackup.sh.erb'),
   }
 
-  if $mysqlbackupdir_target != undef {
+  if $mysqlbackupdir_target {
     file { $backupdir:
       ensure => $mysqlbackupdir_ensure,
       target => $mysqlbackupdir_target,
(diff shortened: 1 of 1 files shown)


tphoney:
Thanks for the PR, @comport3 . Could you add to the documentation, about this great new functionality. 
optional, squash your commits

comport3:
Yes of course. Is the correct file to do this in the CHANGELOG.md ?

Otherwise, here's the new functionality -

Within the class mysql::backup::mysqldump, there is not the ability to change the default behaviour of backing up to a 'directory' to a 'symlink' instead.

The specific use case it solved for us was to ensure backups were stored on a LUKS/DM Crypt based encrypted partition, but could also be handy for environments using NFS shared and similar redirection.

Usage (Hiera)
```
mysql::backup::mysqldump::mysqlbackupdir_ensure: link
mysql::backup::mysqldump::mysqlbackupdir_target: /mnt/mysqldumps
```

tphoney:
Normally it would go into the top of the file. But because this is a private class, there is no docs in the file. For this to be useful for all users, https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/server/backup.pp should be updated to include these new parameters. As it would be useful for all of the backup providers.

tphoney:
Thanks for the pr @comport3 :)

PR #1217, opened 2019-07-29
 Move requires in installdb from log-error file to mysql_datadir

rick-pri:
On FreeBSD I found that I had to ensure that the `log-error` file was in place before the `mysql_datadir` could be run.  The dependency was in reverse but I don't really see a need for that to be the case.  Maybe someone could test with other OSes but for us this fixed the issue where we couldn't get this module to work with Mysql Server 5.7 on FreeBSD 12p8

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -27,7 +27,6 @@ class mysql::server::installdb {
       owner   => $mysqluser,
       group   => $::mysql::server::mysql_group,
       mode    => 'u+rw',
-      require => Mysql_datadir[ $datadir ],
     }
   }
 
@@ -38,6 +37,7 @@ class mysql::server::installdb {
       user                => $mysqluser,
       log_error           => $log_error,
       defaults_extra_file => $_config_file,
+      require             => File[$options['mysqld']['log-error']],
     }
 
     if $mysql::server::restart {


manifests/server/installdb.pp
```diff
@@ -27,7 +27,6 @@
       owner   => $mysqluser,
       group   => $::mysql::server::mysql_group,
       mode    => 'u+rw',
-      require => Mysql_datadir[ $datadir ],
```
smortex: 1/ The relationship between the error log file and the datadir used to exist only if `$options['mysqld']['log-error']` is evaluated to true (condition on line 24)

3/ Possible fix: replace `require` with before (since you change the direction of the relationship)

manifests/server/installdb.pp
```diff
@@ -38,6 +37,7 @@
       user                => $mysqluser,
       log_error           => $log_error,
       defaults_extra_file => $_config_file,
+      require             => File[$options['mysqld']['log-error']],
```
smortex: 2/ This `mysql_datadir` resource is always declared with a relationship, so we might end-up with:

```
require => File[undef]
```

when `$options['mysqld']['log-error']` is not set, which seems to trigger your error.

manifests/server/installdb.pp
```diff
@@ -27,7 +27,6 @@
       owner   => $mysqluser,
       group   => $::mysql::server::mysql_group,
       mode    => 'u+rw',
-      require => Mysql_datadir[ $datadir ],
     }
```
smortex: 4/ Alternate fix: add an arrow dependency in the if condition:

```
File[$options['mysqld']['log-error']] -> Mysql_datadir[ $datadir ]
```

manifests/server/installdb.pp
rick-pri: This what I went with in the end, a `before` rather than `require` on a new branch (since I made a mess of the old one trying to chase down this bug https://tickets.puppetlabs.com/browse/MODULES-3264).  I don't think that I can reopen this one.

PR #1221, opened 2019-07-31
Fix/fix dependency issue in freebsd with log error file creation from 10.0.0

rick-pri:
Change relationship between the log-error file and the mysql datadir.  In FreeBSD this file needs to be present before trying to create the datadir and running the initdb step but will only be created if the option exists.

We should probably also have a require on the log-error file resource for `Package[ $mysql::server::package_name ]` so that the `user` and `group` are present, however, this is probably ensured by the chaining elsewhere.

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -23,11 +23,11 @@ class mysql::server::installdb {
 
   if $options['mysqld']['log-error'] {
     file { $options['mysqld']['log-error']:
-      ensure  => present,
-      owner   => $mysqluser,
-      group   => $::mysql::server::mysql_group,
-      mode    => 'u+rw',
-      require => Mysql_datadir[ $datadir ],
+      ensure => present,
+      owner  => $mysqluser,
+      group  => $::mysql::server::mysql_group,
+      mode   => 'u+rw',
+      before => Mysql_datadir[ $datadir ],
     }
   }
 


smortex:
You have two unrelated commit in the PR (the 2 last).

I think you can fix this situation by:
```
git reset --hard 6fbd5a8
git push -f
```

smortex:
Regarding dependency on user, group, the [file resource will autorequire them](https://puppet.com/docs/puppet/6.7/types/file.html#file-description).

rick-pri:
> You have two unrelated commit in the PR (the 2 last).
> 
> I think you can fix this situation by:
> 
> ```
> git reset --hard 6fbd5a8
> git push -f
> ```

Thanks for that, that's done.


> Regarding dependency on user, group, the [file resource will autorequire them](https://puppet.com/docs/puppet/6.7/types/file.html#file-description).

I know that puppet normally does a good job of autorequiring things like this, however, I think that the user and group are created by the package and that's outside of the realm of what puppet knows about.
Quite often these service accounts created through the packages will actually have things like nologin set as the shell, and a specific home dir, which puppet wouldn't setup with an autorequired user account and I don't know what the package does if it encounters an existing account.

smortex:
> I know that puppet normally does a good job of autorequiring things like this, however, I think that the user and group are created by the package and that's outside of the realm of what puppet knows about.
Oh, I see…  Yes, that totally makes sense!

rick-pri:
> As you said, the user and group will be created by the package, so we should require it.

I don't think that this is required.  The chaining in the `manifests/server.pp` ensures that the package is installed before this manifest.

Here's the full chain out of the `server.pp` for reference:
```
  Anchor['mysql::server::start']
  -> Class['mysql::server::config']
  -> Class['mysql::server::install']
  -> Class['mysql::server::binarylog']
  -> Class['mysql::server::installdb']
  -> Class['mysql::server::service']
  -> Class['mysql::server::root_password']
  -> Class['mysql::server::providers']
-> Anchor['mysql::server::end']
```

PR #1222, opened 2019-08-06
Fix errors for /bin/sh with the xtrabackup cron

baldurmen:
Using Debian's default shell for cron (`/bin/sh`), the xtrabackup incremental cron was failing for me with an error message like:

```
Cron <root@foo> /usr/local/sbin/xtrabackup.sh --incremental-basedir=/foo/mysql-backups --target-dir=/foo/mysql-backups/`date +
/bin/sh: 1: Syntax error: EOF in backquote substitution
```

This is caused by  percent-signs (`%`) not being escaped. With `/bin/sh`, percent-signs (`%`) in the command, unless escaped with backslash \, will be changed into newline characters, and all data after the first % will be sent to the command as standard input. Thus the EOF error.

I also made another small commit to convert backtiks (`) into $(), which is the prefered POSIX way. https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03 has a good writeup on it.

Happy to make modifications if required.

### DIFF
--- a/manifests/backup/xtrabackup.pp
+++ b/manifests/backup/xtrabackup.pp
@@ -63,7 +63,7 @@ class mysql::backup::xtrabackup (
 
   $daily_cron_data = ($incremental_backups) ? {
     true  => {
-      'directories' => "--incremental-basedir=${backupdir} --target-dir=${backupdir}/$(date +%F_%H-%M-%S)",
+      'directories' => "--incremental-basedir=${backupdir} --target-dir=${backupdir}/$(date +\\%F_\\%H-\\%M-\\%S)",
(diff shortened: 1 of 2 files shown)


baldurmen:
Now with updated spec tests.

baldurmen:
Well a bunch of tests fail because GitHub is unstable and cloning repositories fail. Could someone with permissions on this project restart the Travis jobs for me?

baldurmen:
Well, I forced Travis to run again, sorry for closing/reopening but that was the only solution I had.

This seems ready for review :)

PR #1224, opened 2019-08-14
Don't run fact when you can't find mysqld

jstewart612:


jstewart612:
imo don't expose a fact that needs a binary to be shelled out to unless the binary can be located.

PR #1233, opened 2019-09-24
change split on whitespace to split on tab in mysql_user

koshatul:
#### What change does this introduce?

```ruby
max_user_connections, @max_connections_per_hour, @max_queries_per_hour,
      @max_updates_per_hour, ssl_type, ssl_cipher, x509_issuer, x509_subject,
      @password, @plugin = mysql_caller(query, 'regular').split(%r{\s})
```
splits on whitespace, so if the issuer or subject for the user certificate contains spaces it will break all following fields.

#### Why make this change?

When using certificate subjects or issuers with spaces, the split fails and breaks the parsing of the user from MySQL.

puppet code:
```
  mysql_user { "koshatul@%":
    ensure      => 'present',
    tls_options => [
      "SUBJECT /OU=Users/CN=koshatul",
      "ISSUER /O=Acme Widgets/OU=Users/CN=Intermediate Certificate Authority",
    ],
  }
```
will show this notice on every single puppet report.
```
Notice: /Stage[main]/Testing::Mysql/Testing::Mysql::Userwithssl[koshatul]/Mysql_user[koshatul@%]/tls_options: tls_options changed ['ISSUER /O=Acme', 'SUBJECT Widgets/OU=Users/CN=Intermediate'] to ['SUBJECT /OU=Users/CN=koshatul', 'ISSUER /O=Acme Widgets/OU=Users/CN=Intermediate Certificate Authority'] (corrective)
```
#### What approach will be taken?

Change `\s` to `\t`.
the command is run with `-B` which adds `--silent`.
>  -s, --silent        Be more silent. Print results with a tab as separator,
>                      each row on new line.


koshatul:
I don't think any of the travis failures are related to my change.

koshatul:
I'll see if I can find if/when the output changed for the mysql command line.

koshatul:
The documentation on mysql's website only goes back as far as 5.5, it has tab separation for the batch-mode output.

https://dev.mysql.com/doc/refman/5.5/en/mysql-command-options.html#option_mysql_batch

>--batch, -B
>
>Print results using tab as the column separator, with each row on a new line. With this option, mysql does not use the history file.
>
>Batch mode results in nontabular output format and escaping of special characters. Escaping may be disabled by using raw mode; see the description for the --raw option.

The spec test seems to have been spaces since it was refactored from `database_user`, so I think this is actually broken in the real world, it's just lucky that the commonly used fields rarely if ever have spaces.

Surely I can't be the first person to have issues with the TLS config though (unless I'm also the first to use a TLS CA with spaces in the name).

carabasdaniel:
Hi @koshatul 

Thank you for submitting this PR.

I've added a commit to fix the unit testing for this change, however it appears there are some acceptance testing failures caused by this change that you can verify in travis. 

Can you verify and fix the testing for those failures ? You might be able to run the tests on your local machine with Litmus to have an easier time getting these issues solved. Please let us know if you are having trouble with this.

david22swan:
@koshatul Hey, sorry to bother you but I'm just checking to see whether or not you intended to return to this work at any point in time?

koshatul:
@david22swan sorry, been a bit busy lately.

I merged back in master and fixed the issue, I'm about to head off for the day so I'll check later or tomorrow if the tests passed.

koshatul:
I'm writing some acceptance tests for it, but I've noticed another issue is that `ISSUER` and `SUBJECT` require a quoted parameter.

https://dev.mysql.com/doc/refman/5.7/en/create-user.html
```
tls_option: {
   SSL
 | X509
 | CIPHER 'cipher'
 | ISSUER 'issuer'
 | SUBJECT 'subject'
}
```

but tls_options returns unquoted options, so the `idempotent_apply` will fail in acceptance, but there's no test for it at the moment.

koshatul:
Not sure how to acknowledge or complete the change request, but I've added acceptance and spec tests and I've actually pointed the Puppetfile in my local environments at my fork to test it, so it also works for me.

david22swan:
@koshatul Looking over this I can't see one thing wrong so I'm gonna go ahead and merge. :)
Thanks for putting in the work

PR #1207, opened 2019-06-14
Make backup success file path configurable

HT43-bqxFqB:
Using `/tmp` is fine until a node is rebooted and `/tmp` is cleared. By offering a configurable parameter, one can change the path to a non temporary one for checks, monitoring etc.

Additionally I've added a unit test for the default case. Unfortunately the test did not work on my local machine.

PR #1242, opened 2019-10-14
Fix xtrabackup regression introduced in #1207

fraenki:
Currently the `mysql::backup::xtrabackup` class is nonfunctional:

```
Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, 
Class[Mysql::Backup::Xtrabackup]: has no parameter named 'backup_success_file_path' 
(file: /etc/puppetlabs/code/environments/development/modules/mysql/manifests/server/backup.pp, line: 98, column: 3) 
```

This is due to an oversight in #1207, it was forgotten to add the parameter `$backup_success_file_path` to the manifest (but it was added to the template).

This PR adds the missing parameter `$backup_success_file_path`.


fraenki:
The failed test...

```
Error: Execution of '/usr/bin/mysql --database=mysql -e SET PASSWORD FOR 'root'@'localhost' = '*2470C0C06DEE42FD1618BB99005ADCA2EC9D1E19'' returned 1:
```

...is not caused by this PR :)

PR #1243, opened 2019-10-14
Fix mysql::sql task error message

alexjfisher:
Because of misplaced `"` no interpolation of the error message was
actually taking place.  After fixing this, the `Lint/UselessAssignment`
rubocop violation that had been disabled goes away too.


adreyer:
We also need to update the response to add a `_error` hash instead of or in addition to the `error` string so that error messages will be displayed correctly by task runners.  Do you want to do that as part of this PR or should I put one up?

alexjfisher:
@adreyer Would you like to?  I'm not too confident with bolt at the moment.  Similar fix needed in puppetlabs/postgresql (and probably other modules) too.

alexjfisher:
@nicklewis I've found a similar bug in the other task `export.rb`.  Do you want to request a review from @hunner too? ;)

PR #1244, opened 2019-10-15
(MODULES-10023) expose $incremental_backups parameter

fraenki:
https://tickets.puppetlabs.com/browse/MODULES-10023

In #1188 a new parameter `$incremental_backups` was added to the private class `mysql::backup::xtrabackup`.

However, in order to be actually useful and for more convenience this parameter should also be added to the parent class `mysql::server::backup` from which the private class is usually called with the appropiate parameters.

### DIFF
--- a/manifests/backup/mysqlbackup.pp
+++ b/manifests/backup/mysqlbackup.pp
@@ -27,6 +27,7 @@ class mysql::backup::mysqlbackup (
   $postscript               = false,
   $execpath                 = '/usr/bin:/usr/sbin:/bin:/sbin',
   $optional_args            = [],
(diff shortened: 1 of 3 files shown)


lionce:
Hello @fraenki ,

Thank you for this contribution. It would be possible to add a test under [mysql_backup_xtrabackup_spec](https://github.com/puppetlabs/puppetlabs-mysql/blob/505290677d557f817f604b1aca920945a11e3817/spec/classes/mysql_backup_xtrabackup_spec.rb#L161) to test the new param please? Thanks!

Cheers!

fraenki:
This fix was incooporated into #1245.

PR #1247, opened 2019-10-23
Improvement to mysql::sql task

alexjfisher:
* Added a new `structured` parameter.  When `true` output from the mysql
will be parsed by the task and returned as structured data.
* MySQL errors are parsed and detailed error objects returned.

### DIFF
--- a/tasks/sql.json
+++ b/tasks/sql.json
@@ -17,6 +17,10 @@
      "sql": {
       "description": "The SQL you want to execute",
       "type": "String[1]"
+    },
+     "structured": {
+      "description": "Whether you want the returned data parsed for you",
+      "type": "Optional[Boolean]"
     }
   }
 }
--- a/tasks/sql.rb
+++ b/tasks/sql.rb
@@ -3,27 +3,67 @@ require 'json'
 require 'open3'
 require 'puppet'
 
-def get(sql, database, user, password)
-  cmd = ['mysql', '-e', "#{sql} "]
-  cmd << "--database=#{database}" unless database.nil?
-  cmd << "--user=#{user}" unless user.nil?
-  cmd << "--password=#{password}" unless password.nil?
-  stdout, stderr, status = Open3.capture3(*cmd)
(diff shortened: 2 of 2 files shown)


alexjfisher:
Thanks for reviewing this.

I would have liked to have also switched to using[ `puppetlabs-ruby_task_helper`](https://github.com/puppetlabs/puppetlabs-ruby_task_helper).  This would have made adding [unit tests](https://github.com/puppetlabs/puppetlabs-ruby_task_helper#testing) somewhat easier.  As I understand it, this would be a breaking change though as only more modern versions of bolt support  [helper files](https://puppet.com/docs/bolt/latest/writing_tasks.html#sharing-task-code).

Quoting @adreyer from slack.
> Switching to the task helper is a breaking change since multi-file tasks were not supported in PE before 2019.0.1 . I’m not sure if the tasks are considered stable parts of those modules however.

I'll see what I can do.  Maybe I can add some testing regardless.

cmccrisken-puppet:
Hi @alexjfisher - appreciate you looking in to it. If it's something that will cause a time sink, then we can probably merge without the tests. Still in a bit of a learning phase at the moment :)

david22swan:
@alexjfisher Hey, just wondering if this was work that you intended to continue and if there was any help that you needed?

sheenaajay:
@alexjfisher Just wondering is there any more work on this PR. Looks like its ready to merge. Thanks for all your work.

michaeltlombardi:
Since this hasn't seen any activity in over a month I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed, new information is available, or you have a question related to this pull request. We've become aware of difficulties re-opening pull requests, in the event you cannot please mention `puppetlabs/modules`  with an @ sign in front and we'll re-open this pull request.

  Closing the pull request doesn't mean we don't consider this change valuable, just that there are things that need to be addressed before it can be merged. If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

PR #1256, opened 2019-11-18
(MODULES-10114) Confine fact for only when mysql is in PATH

bFekete:
The custom fact should only run when mysql can be found within the PATH environment variable.


florindragos:
Hey @bFekete 
Thanks for your contribution.
Fix looks good!

PR #1258, opened 2019-11-27
Unconditionally set show_diff to false for .my.cnf (remove puppet3 workaround)

smortex:
The show_diff parameter of the file resource was added in Puppet 3.0.
At that time, the module supported previous versions of Puppet so only
set it to false if Puppet was ≥ 3.0.

As of today, the mysql module require Puppet ≥ 5.5.10, so all supported
versions of Puppet support the show_diff parameter.

Remove the conditional logic and always set show_diff to false.


DavidS:
While it's true that we don't commercially support puppet3 any more, I'd rather not commit breaking changes for puppet3 unless there is a overriding technical necessity.

PR #1259, opened 2019-11-27
Fix test suite when mysqld is not installed

smortex:
The fact being confined, no value was returned when mysqld was not
found.  Also stub the which method to fix the problem.


carabasdaniel:
Hi @smortex, thank you for your contribution on fixing this test.

PR #1260, opened 2019-11-29
create the folder for the logfile, not the logfile itself

aloosnetmatch:
i have the manifest for my server:

```
  -> class { '::mysql::server':
    package_name            => 'mariadb-server',
    package_ensure          => latest,
    remove_default_accounts => true,
    root_password           => '<password>',
    restart                 => true,
    override_options        => {
      mysqld => {
        log-error => '/var/log/mariadb/mariadb.log',
        pid-file  => '/var/run/mariadb/mariadb.pid',
      },
      mysqld_safe => {
        log-error => '/var/log/mariadb/mariadb.log',
      },
    }
  }
```

and when running the puppet agent i get this error

```
Error: Could not set 'present' on ensure: No such file or directory @ rb_sysopen - /var/log/mariadb/mariadb.log (file: /etc/puppetlabs/code/environments/linuxtesting/modules/mysql/manifests/server/installdb.pp, line: 25)
Error: Could not set 'present' on ensure: No such file or directory @ rb_sysopen - /var/log/mariadb/mariadb.log (file: /etc/puppetlabs/code/environments/linuxtesting/modules/mysql/manifests/server/installdb.pp, line: 25)
```

It seems like the file "/var/log/mariadb/mariadb.log" cannot be created.(because the folder "/var/log/mariadb/" does not exist.

But that is not what we want, we want to create the folder "/var/log/mariadb/"
so in "installdb" i now use the "dirname" to get the foldername instead of the filename

`file { dirname($options['mysqld']['pid-file']):`

### DIFF
--- a/manifests/server/installdb.pp
+++ b/manifests/server/installdb.pp
@@ -1,4 +1,4 @@
-# @summary 
+# @summary
 #   Builds initial databases on installation.
 #
 # @api private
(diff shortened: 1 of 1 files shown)


sheenaajay:
Thanks @aloosnetmatch for submitting the PR.
Could you please look into the unit test case failures. Thank you.

michaeltlombardi:
Since this hasn't seen any activity in about three months I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed, new information is available, or you have a question related to this pull request. We've become aware of difficulties re-opening pull requests, in the event you cannot please mention `puppetlabs/modules`  with an @ sign in front and we'll re-open this pull request.

Closing the pull request doesn't mean we don't consider this change valuable, just that there are things that need to be addressed before it can be merged. If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

bschonec:
Can we revisit this, please?  I'm having the same problem with MariaDB Enterprise and the lack of /var/log/mariadb is causing the module to fail.

PR #1268, opened 2019-12-10
Use provision job that excludes rpm --rebuilddb command

cmccrisken-puppet:
## Description
Encountering an issue with CentOS 8 hosts failing to provision due to this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1680124

The `rpm --rebuilddb` command was introduced as there were  issues attempting to provision CentOS 6 hosts recently and the introduction of that command seem to resolve that issue: https://github.com/puppetlabs/provision/pull/85

Given that workaround, we will not want to remove this command as there are more CentOS 6 hosts in use than CentOS 8. In fact, from what I can see, `puppetlabs-mysql` is the only module using CentOS 8.

See also: https://github.com/puppetlabs/provision/commit/bd4bb1632bfe0ec24fb7a206910e0e9ad00a7a43


cmccrisken-puppet:
Although this issue was encountered locally, it's not the issue affecting the Travis CI tests.

PR #1206, opened 2019-06-12
Fix regression introduced by PR#1175 for managing CRON package

comport3:
When using the 'puppet-cron' module, the following error was observed when testing release v9.1.0:
`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Package[cron] is already declared at (file: /etc/puppetlabs/code/environments/production/modules/mysql/manifests/backup/mysqldump.pp, line: 67); cannot redeclare (file: /etc/puppetlabs/code/environments/production/modules/cron/manifests/install.pp, line: 10) (file: /etc/puppetlabs/code/environments/production/modules/cron/manifests/install.pp, line: 10, column: 5) on node`

So in classes in the backup classes the parameter "$manage_package_cron" was added with a default of false to allow a non breaking upgrade, with this being overridable as a top class parameter as needed, eg. using Hiera.

PR #1269, opened 2019-12-10
Use ensure_packages() to install  the cron packages

baldurmen:
By default, `mysql::backup::mysqldump` and `mysql::backup::mysqlbackup` classes try to install cron-related packages.

This is pretty annoying, since it makes Puppet fail if any other package tries to do the same thing.

This PR adds a parameter (default: false) to be able to manage this.

### DIFF
--- a/manifests/backup/mysqlbackup.pp
+++ b/manifests/backup/mysqlbackup.pp
@@ -66,17 +66,11 @@ class mysql::backup::mysqlbackup (
   }
 
   if $::osfamily == 'RedHat' and $::operatingsystemmajrelease == '5' {
-    package {'crontabs':
-      ensure => present,
-    }
+    ensure_packages('crontabs')
   } elsif $::osfamily == 'RedHat' {
-    package {'cronie':
-      ensure => present,
-    }
+    ensure_packages('cronie')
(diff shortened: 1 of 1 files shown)


baldurmen:
Travis CI seems broken, but I'm pretty this PR didn't break it :)

Happy to rebase on something that works if needed.

sheenaajay:
@baldurmen Thanks for submitting the PR. Much appreciated. Can we set the value to true by default so the current behaviour won't change and user can use the following variable($install_cron) to install cron package or not . Hence we won't be introducing backward incompatibility. Thank you.

baldurmen:
I was recently showed the `ensure_package()` function by a colleague and it seems like a much cleaner solution to this problem. I'll have a go a changing this patch when I have some time.

sheenaajay:
@baldurmen We have a similar fix in the following PR.https://github.com/puppetlabs/puppetlabs-mysql/pull/1206
Do you want to take a quick look? Thank you.

baldurmen:
Sorry this took so long ... life^WCovid-19 happened.

sheenaajay:
Thank you  @baldurmen for submitting the PR. Good Job.

PR #1276, opened 2020-01-30
Update README.md Ubuntu example to xenial

johnwarburton:
Hi - thanks for this great forge module. It got me over the line when I was forced to use mariadb

Just reviewing my code, and the service name on xenial is mysqld, not mysql, so I thought I'd take the opportunity to refresh this part of the doc, as I copy/pasted verbatim!

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ Class['mysql::bindings']
 
 #### Optional: Install the MariaDB official repo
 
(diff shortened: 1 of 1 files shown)


johnwarburton:
Thanks @cmccrisken-puppet  - I have been out the past few days and planning to push this today - you beat me to it..

PR #1278, opened 2020-02-06
Allow adapting MySQL configuration file's permissions mode

unki:
Sometimes more restrictive file-permissions on the MySQL configuration file are required.

e.g. in case of a Galera cluster, where you need to store `wsrep_sst_auth` credentials in the configuration. In that specific case it's a rather powerful MySQL user and you do not want the file to be be world-readable then.

It would be great if we can support this with _puppetlabs-mysql_ too.

PS: In respect of #32, where it was changed from a restrictive permission to `0644`.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -6,6 +6,7 @@
 class mysql::params {
 
   $manage_config_file     = true
+  $config_file_mode       = '0644'
   $purge_conf_dir         = false
   $restart                = false
   $root_password          = 'UNSET'
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -11,6 +11,8 @@
 #
 # @param config_file
 #   The location, as a path, of the MySQL configuration file.
+# @param config_file_mode
+#   The MySQL configuration file's permissions mode.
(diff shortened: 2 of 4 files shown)


carabasdaniel:
Hello @unki, 

Thank you for your contribution. 

This seems like a feature that would be very helpful, but it would be great if you could add a test for different permissions.

unki:
Hi @carabasdaniel 

> This seems like a feature that would be very helpful, but it would be great if you could add a test for different permissions.

Thanks for considering this PR!
I've added some test-cases just now.

carabasdaniel:
Hi @unki 
Thanks, merging this PR now.

TwizzyDizzy:
Am I right in assuming, that this basically fixes https://tickets.puppetlabs.com/browse/MODULES-8305? If so: thank you :) If not: meh, but still: thank you :D

Cheers
Thomas

unki:
I would say yes - it can be tackled with this change and #1284. Just that I've left the default-values for the permissions on my.cnf as they were before - 0644, root: root.

PR #1284, opened 2020-02-26
Allow changing the mysql-config-file group-ownership

unki:
With my PR #1278 it's now possible to change the file-permission-mode of the mysql-config-file - _e.g._ `my.cnf`.

If you are trying now to apply a rather strict file-mode like `0600` - in combination with Puppet's default file-ownership `root:root` - it will lead to issues with MariaDB not being able to access the config-file after it has switched the processes to the service-user (_e.g._ `mysql`).

Therefor I would propose additionally using `$mysql::server::mysql_group` as the mysql-config-file's group-ownership. Then restricting access would be easy done with a `0640` file-mode.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -96,6 +96,8 @@ class mysql::params {
       $datadir                 = '/var/lib/mysql'
       $root_group              = 'root'
       $mysql_group             = 'mysql'
+      $mycnf_owner             = undef
+      $mycnf_group             = undef
       $socket                  = '/var/lib/mysql/mysql.sock'
       $ssl_ca                  = '/etc/mysql/cacert.pem'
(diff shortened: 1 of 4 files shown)


michaeltlombardi:
Hey @unki, thanks for this patch! Can you rebase your PR on latest master and repush? I believe that may fix your CI issue.

TwizzyDizzy:
Hey @unki,

thanks also for this... seems like we had the same thoughts. What I don't get, is why you - in your second commit - decided not to use `$mysql::server::root_group` but introduce those new variables for mycnf-ownership. Did using `$mysql::server::root_group` have side effects that made those new variables necessary?

Cheers & thanks in advance :)
Thomas

PR #1293, opened 2020-03-27
Improve differences between generated mysql service id values

ryaner:
This uses the entire mac address then reduces the value to the
acceptable range for mysql.

### DIFF
--- a/lib/facter/mysql_server_id.rb
+++ b/lib/facter/mysql_server_id.rb
@@ -1,5 +1,14 @@
 def mysql_id_get
-  Facter.value(:macaddress).split(':')[2..-1].reduce(0) { |total, value| (total << 6) + value.hex }
+  # Convert the existing mac to an integer
+  macval = Facter.value(:macaddress).delete(':').to_i(16)
+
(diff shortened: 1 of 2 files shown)


ryaner:
Some additional background on the reasoning. We've hit multiple collisions with nodes getting the same server-id inside of the same replication chains. It has been getting more common as the number of nodes has increased.

Examples using the old function

mac => "02:00:0a:32:78:cb",
server-id               = 2834123
mac => "04:fe:0a:32:7a:4b",
server-id               = 2834123

mac => "04:FE:0A:32:79:3B"
server-id               = 2834043
mac => "02:00:0A:32:78:7B"
server-id               = 2834043

The new method is obviously not perfect but using the entire mac int value and filtering down should reduce collisions.

ryaner:
I also ran a test using the old vs new method against 1 million random mac addresses.
Old has 7359 collisions. New has 109 collisions.

daianamezdrea:
Hello @ryaner, thank you for your contribution! This looks good !

PR #1162, opened 2019-02-27
FM 7783 - Test Audit making testing more efficient

lionce:
- Idempotent Function 
- removed acceptance tests which are covered by unit tests
- removed deprecated functions  and attached tests 
- added unit tests for edge cases (increased a bit the test coverage)
- reduced execution time
- removed fact mysql_plugin_spec.rb, locales_spec.rb, 

This effort has been realized by both myself and @Lavinia-Dan 

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/linux/view/_adhoc/view/mysql/job/forge-module_puppetlabs-mysql_intn-sys_full-pe-current-adhoc/PLATFORM=redhat5-64default.a,WORKER_LABEL=beaker/

PR #1298, opened 2020-04-28
Revert "Puppet 4 functions"

alexjfisher:
Reverts puppetlabs/puppetlabs-mysql#1274

A puppet 4 namespaced version of this function already existed.
https://github.com/puppetlabs/puppetlabs-mysql/blob/0a8f0054113c68b60f19e381f9452155b937b31b/lib/puppet/functions/mysql/password.rb

A 4x api shim had also been created but then was removed in https://github.com/puppetlabs/puppetlabs-mysql/pull/1162 *without* also removing the `puppet/parser/functions` copy.  sigh.

I'll create a separate PR to put back the shim and remove the legacy API version.


alexjfisher:
As promised separate PR to remove `puppet/parser/functions` version. https://github.com/puppetlabs/puppetlabs-mysql/pull/1299

alexjfisher:
Suggest applying `maintenance` label to this and https://github.com/puppetlabs/puppetlabs-mysql/pull/1274 so that neither appear in the changelog going forwards.

PR #1299, opened 2020-04-28
Remove legacy (old API) `mysql_password` function

alexjfisher:
There's been a properly namespaced modern `mysql::password` function for
some time.  See https://github.com/puppetlabs/puppetlabs-mysql/pull/1044

When the modern API non-namespaced shim was removed in a68004c35ba09fe508020b65a261529437f3a09e
the `puppet/parser/functions` version still existed so users who hadn't
migrated to `mysql::password` would have reverted to this legacy API version.

This commit finally removes the legacy version, but puts back the shim
for now (to not create a breaking change just yet).

Both namespaced and non-namespaced shim are tested.


oniGino:
running module 10.7.0 getting function not found on mysql::password, with puppet 5.5.20

sazzle2611:
running module 10.8.0 with puppet 5.5.22 and getting error Unknown function: 'mysql::password'

Jimadine:
@sazzle2611 @oniGino Either of you work out what was going on with puppet 5.5..2x and the `mysql::password` function? I am facing the same `Error: Evaluation Error: Unknown function: 'mysql::password'.` issue. (puppet 5.5.22 on ubuntu 18)

sazzle2611:
@Jimadine sorry no I didn't, our issue was resolved in the end, probably after an upgrade but can't remember for sure now

alexjfisher:
@Jimadine My guess is that you're calling the function from a module which doesn't list `puppetlabs/mysql` as a dependency in its `metadata.json`?

From https://github.com/puppetlabs/puppet-specifications/blob/master/language/func-api.md ...
> Note: A module that wants to call a function in another module must either have that module listed as a dependency, or have no dependencies entry at all in its metadata.json. This requirement is enforced at runtime to help users keep metadata for modules up to date as users then simply need to install a module and its dependencies. If a module does not have the correct dependencies listed the runtime will simply not find the function.

This restriction was removed in Puppet 6.0.0.  (It was almost removed in 5.5.x too, but the change [reverted](https://tickets.puppetlabs.com/browse/PUP-6964?focusedCommentId=602832&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-602832) due to a regression.)

Jimadine:
@alexjfisher Thank you! That was indeed the cause of the problem.

PR #1300, opened 2020-05-12
pdksync - (maint) - Pdk Update

david22swan:
(maint) - Pdk Update
pdk version: `1.17.0`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -40,6 +40,10 @@ Style/BlockDelimiters:
   Description: Prefer braces for chaining. Mostly an aesthetical choice. Better to
     be consistent then.
   EnforcedStyle: braces_for_chaining
+Style/BracesAroundHashParameters:
+  Description: Braces are required by Ruby 2.7. Cop removed from RuboCop v0.80.0.
+    See https://github.com/rubocop-hq/rubocop/pull/7643
+  Enabled: true
 Style/ClassAndModuleChildren:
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -40,6 +40,10 @@ Style/BlockDelimiters:
   Description: Prefer braces for chaining. Mostly an aesthetical choice. Better to
     be consistent then.
   EnforcedStyle: braces_for_chaining
+Style/BracesAroundHashParameters:
+  Description: Braces are required by Ruby 2.7. Cop removed from RuboCop v0.80.0.
+    See https://github.com/rubocop-hq/rubocop/pull/7643
+  Enabled: true
```
alexjfisher, line 46: @david22swan How did this end up being `Enabled: true`? I can't work it out given in https://github.com/puppetlabs/pdk-templates/pull/329 it was set to `Enabled: false`.

alexjfisher: Actually, it's not specific to this module.  Looks like all modules have got `Enabled: true`.
I'm really confused...

alexjfisher: @DavidS Given https://tickets.puppetlabs.com/browse/IAC-778, I should probably ping you too.

alexjfisher: I think I've fixed the issue in https://github.com/puppetlabs/pdk-templates/pull/344

PR #1305, opened 2020-06-01
Add managed_dirs parameter

evgenkisel:
These changes resolve a problem described in the issue [MODULES-10679](https://tickets.puppetlabs.com/browse/MODULES-10679).
Updated the class `mysql::server` to use the new parameter `managed_dirs` which allows to set a list of directories which can be managed by the class `mysql::server::managed_dirs`


sheenaajay:
Thank you @evgenkisel for submitting the PR.

manifests/server.pp
```diff
@@ -88,6 +88,7 @@
                   $remove_default_accounts = false,
                   $restart                 = $mysql::params::restart,
                   $root_group              = $mysql::params::root_group,
+                  $managed_dirs            = undef,
```
sheenaajay, line 91: $managed_dirs            = $mysql::params::managed_dirs
Could you please add the variable to params file and replace with this line above. Thank you
https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/params.pp

evgenkisel: @sheenaajay the file https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/params.pp already contains the variable with different values which related with cases.
As I understand you'd like to change a value for the parameter in the class mysql::server
please see my latest commit https://github.com/puppetlabs/puppetlabs-mysql/pull/1305/commits/a22132740b0da0b24eabf70a46dc0db508a849b2

manifests/server/config.pp
```diff
@@ -36,21 +36,26 @@
...
-      if ( $dir and $dir != '/usr' and $dir != '/tmp' ) {
-        exec {"${entry}-managed_dir-mkdir":
-          command => "/bin/mkdir -p ${dir}",
-          unless  => "/usr/bin/dpkg -s ${mysql::server::package_name}",
-          notify  =>  Exec["${entry}-managed_dir-chmod"],
-        }
-        exec {"${entry}-managed_dir-chmod":
-          command     => "/bin/chmod 777 ${dir}",
-          refreshonly => true,
+  case $::operatingsystem {
```
sheenaajay, line 39: Thank you @evgenkisel for incorporating the comments. By adding the operating system specific it will fail on other oses. So we may have to include conditions for other OSes too to avoid regression. Please let us know if you need more information. Thanks for your contribution.

sheenaajay: Got it. This code is supposed to run only on Debian. Perfect. Running the release checks across all OS and verifying the changes. Thanks again.

PR #1307, opened 2020-06-06
Use ALTER USER for mariadb 10.2.0 or newer

mcb30:
As of MariaDB 10.4.1 the mysql.user table has been replaced by a view
onto the mysql.global_priv table, and the "UPDATE mysql.user" syntax
therefore no longer works.  Support for the newer ALTER USER statement
was added in MariaDB 10.2.0.

Fix the plugin attribute accessor to use ALTER USER for mariadb 10.2.0
or newer.  This matches the existing version check logic within the
password_hash and tls_options accessors.

Signed-off-by: Michael Brown <mbrown@fensystems.co.uk>


daianamezdrea:
Hi @mcb30, thank you for your contribution! This looks good! Cheers!

PR #1312, opened 2020-06-30
Remove control character from manifests/server.pp

tomkrouper:
A check for ascii characters revealed that the contraction for "do not"
was not a valid ascii character, this changes that.

resolves: https://tickets.puppetlabs.com/browse/MODULES-10723


sheenaajay:
Thank you @tomkrouper for submitting the PR.

PR #1315, opened 2020-07-21
Remove non printable characters

elmobp:
In the rare event MySQL spits out something non printable this should cleanse it see the error below:

```
Error: Could not prefetch mysql_user provider 'mysql': invalid byte sequence in UTF-8
Error: Failed to apply catalog: invalid byte sequence in UTF-8
```


daianamezdrea:
Hi @elmobp, this looks good! Thank you for your contribution!

PR #1335, opened 2020-10-14
Do not set empty password as policy don't allow that

scornelissen85:
Mysql server doesn't allow setting an empty password. With this PR the root password would be directly set by mysqladmin. Also /root/.my.cnf is first set before the root user is managed. Without this ordering, mysql_user() cannot be used as the root password has a password set, which isn't set in the client config.

### DIFF
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -13,8 +13,8 @@ class mysql::server::root_password {
   # below exec will remove this default password. If the user has supplied a root
   # password it will be set further down with the mysql_user resource.
   $rm_pass_cmd = join([
-      "mysqladmin -u root --password=\$(grep -o '[^ ]\\+\$' ${secret_file}) password ''",
-      "rm -f ${secret_file}",
+      "mysqladmin -u root --password=\$(grep -o '[^ ]\\+\$' ${secret_file}) password \"${mysql::server::root_password}\"",
+      "rm -f ${secret_file}"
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
It looks like there's some fairly extensive [test failures](https://travis-ci.org/github/puppetlabs/puppetlabs-mysql/jobs/735778067) related to the [unit tests for `mysql::server::root_password`]((https://github.com/puppetlabs/puppetlabs-mysql/blob/main/spec/classes/mysql_server_spec.rb#L155) - updating default behavior can have some knock-on effects, so ensuring that these unit tests (and the acceptance tests that won't run til they clear) pass will help enormously.

Please update the tests to reflect your changes here - if you have any questions or hit any roadblocks, don't hesitate to reach out here or in slack!

pmcmaw:
Hey @scornelissen85 just wondering if this is something you plan on revisiting?

Many thanks
Paula

scornelissen85:
Currently no, I did see later on that it was not easy to implement for an universal solution. When I'm working on that I will create a new PR.

pmcmaw:
Thats no problem, I would just like to thank for taking the time to implement the initial functionality. As there is no plan for a revisit I am going to close this PR. If you disagree with my decision please feel free to let me know. This PR may also be reopened at a later date if you decide to continue working on the test failures. 

Many thanks
Puppet IAC Team

PR #1348, opened 2020-12-04
Repair check of logbindir

qha:
Presumably the intention was to check that it is *both* not relative
and not already managed.

After update to 10.8 the mysql module tries and fails to manage '.'
on nodes where i have log-bin set to a relative path.

### DIFF
--- a/manifests/server/managed_dirs.pp
+++ b/manifests/server/managed_dirs.pp
@@ -23,6 +23,8 @@ class mysql::server::managed_dirs {
         }
       }
     }
+  } else {
+    $managed_dirs_path = []
   }
 
   $logbin = pick($options['mysqld']['log-bin'], $options['mysqld']['log_bin'], false)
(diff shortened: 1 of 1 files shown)


gnif:
I can confirm this issue is valid and the fix is correct. I just hit the identical issue.

qha:
I note that checks fail, i would be willing to put some time into repairing them if someone would provide some starting pointers.

david22swan:
Closing and re-opening pr in order to force the rekick of the tests. Some of the failures look like they may come from travis

david22swan:
@qha Taking a quick look at the failures, quite a few seem to come from some missing changes.
Could you rebase your changes onto the main, that should clear some of them up and make diagnosis easier.

qha:
Rebased, hope that does the trick. Thank you for your time and advice! @david22swan

qha:
Finally figured out at least one reason the tests were failing that was actually a bug (i don't think i introduced it, just changed things up so the tests caught it).

The remaining errors seem to be test environment problems if i understand the output...

david22swan:
@qha Thanks for responding so fast with the changes 👍 
I did a few re-kicks of the tests and their all passing now so I'm happy to merge

qha:
Thanks for merging! Do you know when a new release might happen? @david22swan

qha:
I see that a new version has been released now.

pmcmaw:
Yes apologies, this module has been released as of today and is now available on the Forge! 
Thank you for your patience :-) @qha

PR #1355, opened 2020-12-25
SQL Log bin support for User, Grant, Database.

ebogdanov:
Hi.

I do not expect this to be merged into main, just share solution with community.
This is updated to current version patch of https://github.com/puppetlabs/puppetlabs-mysql/pull/894/files

If you have problem with errant transaction, after creation users/grants on slaves, like analytical or something like that, you can use this approach.

### DIFF
--- a/examples/mysql_user.pp
+++ b/examples/mysql_user.pp
@@ -23,6 +23,7 @@ mysql_user { 'dan@%':
 mysql_user { 'socketplugin@%':
   ensure => present,
   plugin => 'unix_socket',
+  bin_log => 'no',
 }
 
 mysql_user { 'socketplugin@%':
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -99,21 +99,37 @@ class Puppet::Provider::Mysql < Puppet::Provider
     self.class.defaults_file
   end
 
-  def self.mysql_caller(text_of_sql, type)
+  def self.mysql_caller(text_of_sql, type, bin_log = 'yes')
+    opt = []
     if type.eql? 'system'
       if File.file?("#{Facter.value(:root_home)}/.mylogin.cnf")
         ENV['MYSQL_TEST_LOGIN_FILE'] = "#{Facter.value(:root_home)}/.mylogin.cnf"
-        mysql_raw([system_database, '-e', text_of_sql].flatten.compact).scrub
       else
-        mysql_raw([defaults_file, system_database, '-e', text_of_sql].flatten.compact).scrub
+        opt.push(defaults_file)
       end
+      opt.push(system_database)
+
+      if bin_log.eql? 'no'
+        opt.push("--init-command=SET SESSION SQL_LOG_BIN = 0;")
+      end
+      opt.push('-e')
+      opt.push(text_of_sql)
+
+      mysql_raw(opt.flatten.compact).scrub
     elsif type.eql? 'regular'
       if File.file?("#{Facter.value(:root_home)}/.mylogin.cnf")
         ENV['MYSQL_TEST_LOGIN_FILE'] = "#{Facter.value(:root_home)}/.mylogin.cnf"
-        mysql_raw(['-NBe', text_of_sql].flatten.compact).scrub
       else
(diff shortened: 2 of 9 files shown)


gdubicki:
I praise the inititive but it doesn't work for me, @ebogdanov . Is it working for you?

Firstly I had to add something like:
```
  newproperty(:bin_log) do
    desc 'Indicates if SQL_LOG_BIN should be used for this user. Helpful for slaves.'
    newvalue(%r{\w*})
  end
```
...to `mysql_user.rb` and `mysql_grant.rb` to avoid errors like "Error 500 on SERVER: Server Error: no parameter named 'bin_log'".

Now I am struggling with a problem that although I did set `bin_log => 'no'` in my puppet code the user and grant that I created did cause errant ids and broke replication...

ebogdanov:
Hi @gdubicki 
Yes, this works for me.

Sorry for any inconvenience.

You can restore replication with next workaround:
> stop slave;
> set global gtid_mode=ON_PERMISSIVE;
> SET GLOBAL SQL_SLAVE_SKIP_COUNTER = 1;
> START SLAVE;
> SHOW SLAVE STATUS\G;
> 
> set global gtid_mode=ON;

or use tools described in this Percona's article: https://www.percona.com/blog/2015/12/02/gtid-failover-with-mysqlslavetrx-fix-errant-transactions/

Can you check 2 next things, please:

1.Ensure that this transaction broken replication in your case. I've checked recently case when puppet for some reason decided to execute "alter password" for already existing user, which had not "bin_log" specified, so that errant was created.

Easiest way is to scan bin logs for transaction with this errant gtid, something like: 

> mysqlbinlog --base64_output=DECODE-ROWS --verbose mysql-bin.* | grep "0d9092c6-1f5c-11eb-81c3-9a72c43eddb0:143" -A10

2. sudo puppet agent -tv | grep 'your user'
This will show which commands are executed in your case, if in step 1. you got the transaction which you're mentioned originally.

gdubicki:
Hey @ebogdanov , thanks for a quick response!

I managed to make it work after 3 changes:

1. I change the property bin_log in the types to sth like:
```
newproperty(:bin_log) do
    desc 'Indicates if SQL_LOG_BIN should be used for this user. Helpful for slaves.'
    # This property is used only during resource creation
    # so changing its value should not trigger resource update (correction)
    def insync?(is)
      true
    end
  end
```

...as without this change I had constant info about "corrective" changes of these resources.

2. I changed grant's flush to make it use bin_log property too:

```
  def flush
    @property_hash.clear
    self.class.mysql_caller('FLUSH PRIVILEGES', 'regular', @resource[:bin_log])
  end
```

3. It seems that it was necessary for me to add quoting to the command passes to `--init-command`:

```
opt.push('--init-command="SET SESSION SQL_LOG_BIN = 0;"')
```

...but because of that I had to switch from using a command to `Puppet::Util::Execution` to actually run MySQL cli as I got into quoting hell. :/ 

I have a working solution now and I will share it fully here (maybe as a PR to your repo?) after I polish some rough edges.

gdubicki:
I created a PR with our changes: https://github.com/ebogdanov/puppetlabs-mysql/pull/1

Sorry for the most probably lame code but Ruby is not my native language. :P 

Note that I would love to get back from using `Puppet::Util::Execution` to standard command but I didn't know how...

pmcmaw:
Hey @ebogdanov 

I have had a quick glance at your PR and it looks like a great feature we would be happy with integrating into this code base. If this is something you would be interested in, please feel free to continue working on and we will work on getting it merged. 

Please note the following PR https://github.com/puppetlabs/puppetlabs-mysql/pull/894/files is a feature the team would have been happy to merge, however it was closed due to inactivity and the actual feature would have been great to have. 

Many Thanks 
The IAC Team :-)

pmcmaw:
Hey @ebogdanov I would like to thank you for your time submitting this PR. However as it hasn't made any movement for some time I am going to close this. I apologize and if you disagree with my decision, please feel free to reopen. From comments however it seems that this PR has just been create as a point of reference. 

If you need anything more from my side, please do not hesitate to reach out. 

Many Thanks 
Puppet IAC Team. :-)

PR #1357, opened 2021-01-07
(IAC-1230) Bring inline with other examples

pmcmaw:
Currently other examples are using _ instead of :: so bringing this example inline with the others.


pmcmaw:
This should not be change this is not correcting what I thought it should be doing.

PR #1360, opened 2021-01-16
Set default MySQL version for FreeBSD

olevole:
FreeBSD uses MySQL 57 by default
Ref: https://svnweb.freebsd.org/ports/branches/2021Q1/Mk/bsd.default-versions.mk?revision=560000&view=markup#l83

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -328,8 +328,8 @@ class mysql::params {
     }
 
     'FreeBSD': {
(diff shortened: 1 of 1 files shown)


david22swan:
@olevole 
Thanks for the work you've put in.
The changes you've made look good to me so I've gone ahead and merged it.
Hope to see more from you in the future.

PR #1378, opened 2021-03-31
Fix: Puppet Unknown variable: 'mysql::params::exec_path'

JvGinkel:


manifests/db.pp
```diff
@@ -53,14 +53,20 @@
   Enum['absent', 'present'] $ensure           = 'present',
   $import_timeout                             = 300,
   $import_cat_cmd                             = 'cat',
-  $mysql_exec_path                            = $mysql::params::exec_path,
+  $mysql_exec_path                            = undef,
 ) {
```
adrianiurca, line 57: To fetch all the parameters from mysql::params you should inherit the params class. Use this line to do this:
`inherits mysql::params`. If you use this you will don't need to verify if `$mysql_exec_path` is defined or not.
Please try it and let us know if it's working for you

adrianiurca: Here is an example of how to use inheritance in puppet https://github.com/puppetlabs/puppetlabs-mysql/blob/bb9e0fcb4da091d4c3a014e6290b679af2a642d7/manifests/server.pp#L117

JvGinkel: You can't use inherits on a defined type I tried this on https://github.com/puppetlabs/puppetlabs-mysql/pull/1377/files as you can see but the pipeline failed.

adrianiurca: Ohh you're right, please use `include mysql::params`.
Something like this https://puppet.com/docs/puppet/5.5/lang_defined_types.html#containment

JvGinkel: I did not test this, but do you suggest to reverse my changes and only include the `mysql::params` ?  Is this not already being covered by the `include 'mysql::client'` as this class is also including the `mysql::params`?

adrianiurca: Hi @JvGinkel, I tested your changes and it's working! I will proceed to merge this PR.
Thanks for your contributions!

kind regards,
@adrianiurca

adrianiurca (approved):

PR #1388, opened 2021-05-19
(maint) Update epel URL

pmcmaw:
The old URL is no longer available. 
Failures on Ubuntu20.04 are ticketed and known: https://tickets.puppetlabs.com/browse/IAC-1612


pmcmaw:
No longer needed. PR was put up when tests were failing as the download image was not found. This is now passing again so the mirror must have just went to temporarily.

PR #1396, opened 2021-05-31
Do not pull python-mysqldb in Debian Bullseye

thomasgoirand:
The package python-mysqldb doesn't exist in Bullseye anymore.


david22swan:
This looks good to me so I'm going to go ahead and merge.
Only change needed is the PR title for the changelog but I can handle that myself, no offense intended.

Thank you for catching this for us and putting up the change, looking forward to seeing more from you in the future.
Have a great day

PR #1400, opened 2021-06-16
Use Puppet-Datatype Sensitive

cocker-cc:
- fix some Puppetlint-Complaints, f.e. use Hash $::facts['blubb'] instead of specific global Variables
- let Function mysql::password accept Datatype Sensitive for Password
- let Function mysql_password accept Datatype Sensitive for Password
- add Parameter "sensitive" to Function mysql::password to decide if its Returnvalue should be of Datatype Sensitive
- add Parameter "sensitive" to Function mysql_password to decide if its Returnvalue should be of Datatype Sensitive
- let mysql_user accept Datatype Sensitive for  Parameter password_hash
- let mysql::backup::mysqlbackup accept Datatype Sensitive for Parameter $backuppassword
- let mysql::backup::mysqldump accept Datatype Sensitive for Parameter $backuppassword
- let mysql::backup::xtrabackup accept Datatype Sensitive for Parameter $backuppassword
- let mysql::db accept Datatype Sensitive for Parameter $password
- let mysql::server accept Datatype Sensitive for Parameter $root_password
- let mysql::server::backup accept Datatype Sensitive for Parameter $backuppassword
- let mysql::server::monitor accept Datatype Sensitive for Parameter $mysql_monitor_password
- let mysql::server::root_password accept Datatype Sensitive for $root_password
- refactor Logic for unset Password for mysql::server::root_password

### DIFF
--- a/manifests/backup/mysqlbackup.pp
+++ b/manifests/backup/mysqlbackup.pp
@@ -5,7 +5,7 @@
 #
 class mysql::backup::mysqlbackup (
   $backupuser               = '',
-  $backuppassword           = '',
+  Variant[String, Sensitive[String]] $backuppassword = '',
   $maxallowedpacket         = '1M',
   $backupdir                = '',
   $backupdirmode            = '0700',
@@ -32,6 +32,11 @@ class mysql::backup::mysqlbackup (
   $compression_command      = undef,
   $compression_extension    = undef,
 ) inherits mysql::params {
+  $backuppassword_unsensitive = if $backuppassword =~ Sensitive {
+    $backuppassword.unwrap
+  } else {
+    $backuppassword
+  }
   mysql_user { "${backupuser}@localhost":
     ensure        => $ensure,
     password_hash => mysql::password($backuppassword),
@@ -104,7 +109,7 @@ class mysql::backup::mysqlbackup (
       'incremental_base'       => 'history:last_backup',
       'incremental_backup_dir' => $backupdir,
       'user'                   => $backupuser,
-      'password'               => $backuppassword,
+      'password'               => $backuppassword_unsensitive
     },
   }
   $options = mysql::normalise_and_deepmerge($default_options, $mysql::server::override_options)
--- a/spec/functions/mysql_password_spec.rb
+++ b/spec/functions/mysql_password_spec.rb
@@ -11,12 +11,22 @@ shared_examples 'mysql::password function' do
     is_expected.to run.with_params.and_raise_error(ArgumentError)
   end
 
-  it 'raises a ArgumentError if there is more than 1 arguments' do
-    is_expected.to run.with_params('foo', 'bar').and_raise_error(ArgumentError)
+  it 'raises a ArgumentError if there is more than 2 arguments' do
+    is_expected.to run.with_params('foo', false, 'bar').and_raise_error(ArgumentError)
   end
 
   it 'converts password into a hash' do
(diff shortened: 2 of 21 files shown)


spec/functions/mysql_password_spec.rb
```diff
@@ -11,14 +11,23 @@
...
   it 'converts password into a hash' do
     is_expected.to run.with_params('password').and_return('*2470C0C06DEE42FD1618BB99005ADCA2EC9D1E19')
   end
 
+  it 'accept password as Sensitive' do
+    is_expected.to run.with_params(sensitive('password')).and_return('*2470C0C06DEE42FD1618BB99005ADCA2EC9D1E19')
+  end
+
+  # Test of a Returnvalue of Datatype Sensitive does not work:
+  # it 'returns Sensitive with sensitive=true' do
```
pmcmaw, line 27: Hey @cocker-cc 

Just wondering if you have more information on why this test does not work? I would be inclined to advise if there is a suitable reason, marking this test in the suite as pending and just adding the comment as to why for future reference.

manifests/backup/mysqlbackup.pp
```diff
@@ -32,6 +32,11 @@
   $compression_command      = undef,
   $compression_extension    = undef,
 ) inherits mysql::params {
+  if $backuppassword =~ Sensitive {
+    $_backuppassword = $backuppassword.unwrap
```
pmcmaw, line 36: I know this isn't a big deal but would it be possible to rename this variable? I found when reviewing this PR with a colleague it took us a while to work out what it is doing. I would suggest using something such as `plaintext_backuppassword`. (this name change will need revisited throughout your codebase changes).

pmcmaw:
Hey @cocker-cc 
I would like to thank you for such an awesome PR! 
I have added some minor comments and as soon as they are addressed I will work on getting this PR merged.
🥇

spec/functions/mysql_password_spec.rb
pmcmaw: Here is an example from stdlib on how to mark as pending: https://github.com/puppetlabs/puppetlabs-stdlib/blob/276e7b7cd97580c9c08b264b3727e5902df1f487/spec/functions/is_mac_address_spec.rb#L22

cocker-cc: 👍

manifests/backup/mysqlbackup.pp
cocker-cc: Done 👍

cocker-cc: Now it's done.  I needed a second `push`.

cocker-cc:
> I would like to thank you for such an awesome PR!
> 1st_place_medal

Thanks.  Reason for that one is: we switched from Trocla to Vault, and our Function `vault_secrets::lookup` natively returns a `Sensitive`, and that's good. My Understanding of `Sensitive` is, that the `unwrap` should be done as late as possible.  Using EPP instead of ERB, in fact there never has to be an `unwrap`.  Maybe I will spawn another Pullrequest with the relevant EPP-Files.

manifests/backup/mysqlbackup.pp
cocker-cc: https://github.com/cocker-cc/puppetlabs-mysql/compare/f9d507f..8413dbe

pmcmaw:
What a fantastic addition to our code base @cocker-cc. 
Again thank you for such a quick turnaround and such a great feature.

pmcmaw (approved):

pmcmaw:
Changes will be released on our next Community Day (Monday)

PR #1404, opened 2021-06-23
Fix issue that prevents Litmus from working when testing locally

ghoneycutt:


ghoneycutt:
Without this change, the provision step hangs indefinitely and you cannot continue the steps to run acceptance tests.

Also `docker_exp` is not documented anywhere. See https://github.com/puppetlabs/provision/issues/170

PR #1405, opened 2021-06-26
(MODULES-11115) add Rocky Linux 8 compatibility

vchepkov:


ghoneycutt:
Suggest adding this to the `provision.yaml` so acceptance testing shows this platform works.

vchepkov:
I don't mind, just don't know how to find out what images are available.
Vagrant rockylinux/8 is there

pmcmaw:
Hey @vchepkov 

We would be more than happy to accept your PR (the code change) to make this module compatible with Rocky. However when adding to the `metadata.json` and `provision.yaml` it suggests that we would be supporting this OS which is something the team would not be able to do. 

Note - I have updated your PR title to include compatibility rather than support.

vchepkov:
@pmcmaw , would you like me to remove change to metadata.json? As long as module's manifest doesn't make any decision based on the metadata, it should be fine. Some modules do that, unfortunately, https://github.com/voxpupuli/puppet-yum/blob/master/manifests/init.pp#L108

pmcmaw:
Yeah none of our supported modules do that. If you remove the metadata change and just have the manifest change I will get it merged. There was a fix merged earlier for spec test failures so if you rebase they will pass. 👍

vchepkov:
@pmcmaw , done

pmcmaw (approved):

PR #1407, opened 2021-06-29
Remove backup classes

ghoneycutt:
Suggest that we remove these backup related classes. The work is useful and should be kept as example profiles. Backups are extremely site specific and having them in their current state is not providing value. These would do better converted to example profiles to show how you can do backups using these different methods. Moving them to examples also makes the module easier to maintain.


pmcmaw:
We are currently planning work on doing a complete revamp of backups as soon as the team get a chance. See epic here: https://tickets.puppetlabs.com/browse/IAC-1661

ghoneycutt:
@LukasAud Could you please take a look at this one. It is similar in nature to the one you just merged.

alexjfisher:
> @LukasAud Could you please take a look at this one. It is similar in nature to the one you just merged.

Is it? mysql::server::backup is documented in the README, and is not related to unsupported Operating Systems.  There's also https://github.com/puppetlabs/puppetlabs-mysql/pull/1447 open where the author has been asked to rebase so that the change can be reviewed properly before merging.

I'm against removing this functionality which I assume actually has a fair few users and seen willingness from community members to contribute fixes and enhancements.

alexjfisher:
One person's cruft is another's functionality they rely on.  Unlike the solaris code, this is out of the way and not interfering with non users.

fraenki:
> Backups are extremely site specific and having them in their current state is not providing value.

I strongly disagree. They are very useful, at least for our use-cases.

pmcmaw:
Closing in line with above comments.

PR #1408, opened 2021-06-29
Deprecate mysql::server::monitor and show as an example

ghoneycutt:
The `mysql::server::monitor` class should be retired and moved to the examples to show how this is done. This is an incredibly site specific thing and made sense when contributed 10 years ago, though no longer should be an active class.


pmcmaw:
This should go green if you rebase.

ghoneycutt:
These errors are unrelated. The only testing for this code were the unit tests which are removed in this commit.

PR #1413, opened 2021-07-01
(feat) allow to set and run backups using another user via cron

tejasbhosale009:
allow to set and run backups using another user via cron , default is set as root via params if no cronbackupuser is specified

### DIFF
--- a/manifests/server/backup.pp
+++ b/manifests/server/backup.pp
@@ -112,7 +112,7 @@ class mysql::server::backup (
   create_resources('class', {
       "mysql::backup::${provider}" => {
         'backupuser'               => $backupuser,
-	'backupcronuser'           => $backupcronuser,
+        'backupcronuser'           => $backupcronuser,
         'backuppassword'           => $backuppassword,
         'backupdir'                => $backupdir,
         'backupdirmode'            => $backupdirmode,
--- a/manifests/server/backup.pp
+++ b/manifests/server/backup.pp
@@ -18,6 +18,8 @@
 #
 # @param backupuser
 #   MySQL user to create with backup administrator privileges.
+# @param backupcronuser
+#   User to use for backup cron jobs. Defaults to root.
 # @param backuppassword
 #   Password to create for `backupuser`.
 # @param backupdir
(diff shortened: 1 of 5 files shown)


pmcmaw:
@tejasbhosale009 the spec tests are failing due to a minor syntax error. I tried to push the fix to the PR but I don't have the permission. Apologies. If you could look at resolving this we will try to get it over the line and merged into main. :-)

tejasbhosale009:
> @tejasbhosale009 the spec tests are failing due to a minor syntax error. I tried to push the fix to the PR but I don't have the permission. Apologies. If you could look at resolving this we will try to get it over the line and merged into main. :-)

Syntax error fixed ..

tejasbhosale009:
@pmcmaw can you please merge changes if all are ok ..

manifests/server/backup.pp
```diff
@@ -109,10 +112,11 @@
   create_resources('class', {
       "mysql::backup::${provider}" => {
         'backupuser'               => $backupuser,
+        'backupcronuser'           => $backupcronuser,
         'backuppassword'           => $backuppassword,
         'backupdir'                => $backupdir,
         'backupdirmode'            => $backupdirmode,
-        'backupdirowner'           => $backupdirowner,
+        'backupdirowner'           => $backupcronuser,
```
david22swan, line 119: Any reason that you have changed this line?
Seem's like ti should stay as it is.

david22swan:
One comment on your changes, look's like it is causing some unit test failures as well

manifests/server/backup.pp
tejasbhosale009: Yes .. this is directory owner where we keep backups so if this is not set to same user by which we are running cron by default it will be root and you will not able to store backup to that directory location after backups are performed.

LukasAud:
Closing and reopening to trigger tests again.

LukasAud:
Hi @tejasbhosale009, we have recently released a new version of this module. Can you rebase your PR to make sure it's up to date?

Also, as @david22swan mentioned before, there seems to be a number of Unit test failures that need to be addressed before we can think about merging your changes. Please take a look at those and make the necessary changes.

PR #1416, opened 2021-07-05
(maint) Update the since_tag

pmcmaw:
The top entry on the HISTORY.md file will be used as the since_tag if it
is not specified anywhere else. This is in relation to automatic change log generation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 # Change log
 
-All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org).
-
-## [v11.0.3](https://github.com/puppetlabs/puppetlabs-mysql/tree/v11.0.3) (2021-06-21)
+## v11.0.3
 
 [Full Changelog](https://github.com/puppetlabs/puppetlabs-mysql/compare/v11.0.2...v11.0.3)
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,161 @@
-# Change log
+## v11.0.3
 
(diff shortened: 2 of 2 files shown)


HISTORY.md
```diff
@@ -1091,3 +1246,6 @@ configuration variables.
 
 
 \* *This Changelog was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
+
```
david22swan, line 1249: Duplicate Entry

CHANGELOG.md
```diff
@@ -1,8 +1,6 @@
 # Change log
 
-All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org).
```
david22swan, line 3: This doesn't really need changed since it will get taken from the HISTORY.md.
Also, should the link info and release date be added below it? Just to maintain the info?

HISTORY.md
pmcmaw: They're 2 different repos. One is under the namespace skywinder and one github-changelog-generator

CHANGELOG.md
pmcmaw: When running the changelog generator this is what it done I haven't made any manual changes.
As it is automated if I was to add back in the information it will be removed again. 
It is just how it works, in order for it to pick it up I needs to be v11.0.3 with nothing else.

HISTORY.md
david22swan: Ah mb

CHANGELOG.md
david22swan: Didn't think youd already ran it, looks good to me then

david22swan (approved):

PR #1418, opened 2021-07-08
Deprecate unsupported Solaris platform

ghoneycutt:


ghoneycutt:
Solaris is not supported and compatibility should be removed from this module as keeping it in makes it harder to maintain and refactor. There are no acceptance tests. It is not even clear as to which version(s) of Solaris this even works on.

After much discussion, Vox Pupuli has decided that unsupported platforms may have their code removed from modules.
https://voxpupuli.org/docs/software_support_lifecycle/

alexjfisher:
> After much discussion, Vox Pupuli has decided that unsupported platforms may have their code removed from modules.
> https://voxpupuli.org/docs/software_support_lifecycle/

That post talks about EOL OSes where EOL date is "determined by the upstream maintainer* of the operating system."
In the case of Solaris, this is Oracle who will apparently be supporting [Solaris 11 until at least 2034](https://blogs.oracle.com/solaris/post/long-live-solaris-11-until-at-least-2034-to-be-exact).

Given that Solaris 11 is also a [PE supported OS](https://puppet.com/docs/pe/2021.2/supported_operating_systems.html#supported_operating_systems_and_devices-supported-agent-platforms), and this is a Puppet Inc supported module, I'm not sure **removing** (this change doesn't deprecate anything, it removes support completely), Solaris compatibility code is going to go down well with some of Puppet Inc's customers.

All that said... I agree that it's very difficult to know if any of this code actually works at present given the lack of visible acceptance tests for Solaris.  My understanding is that Puppet also have some private Jenkins based test frameworks.  Perhaps this module is tested there??  Perhaps @genebean or @binford2k know more on this (and whether there's anyway for trusted community members to get greater visibility) ?

ghoneycutt:
Thanks @alexjfisher for providing that insight.

If Solaris is meant to be supported, at a bare minimum, the metadata should include the versions so it is explicitly stated.

Ideally there would also be testing provided by Puppet. If there is no testing, Puppet Labs cannot really support the code anyway for clients and we should just remove the code instead of saying the module is compatible as its compatibility cannot even be verified.

daianamezdrea:
Hi @ghoneycutt, @alexjfisher, I want to highlight here the difference between supported and compatible. As you said in a previous comment, solaris is not in the metadata.json file, so it's not supported by this module (it's not tested on solaris), but it might be compatible. Indeed , we had some private Jenkins based test frameworks, but still we didn't have tested against solaris and we don't have images for solaris in our nightlies/test PRs from GCP (what we currently use, we used Jenkins in the past). As a conclusion, solaris is not supported, but compatible (maybe) with this module, so, I'll say not to merge this PR, but @binford2k can have the final answer here.

binford2k:
imo, in the module's current state, this code must be removed.

If there were a section of the README that made it completely clear which platforms were **Supported** and a list of untested user contributed platforms that _might work_ then we could consider leaving it in.

ghoneycutt:
@binford2k I think to accommodate @daianamezdrea ideas on compatibility this PR while removing Solaris, should also remove the fail by default and change that to something like `warn('Detected platform is not supported')`. In this way the module can still be used on Solaris or other unsupported yet possibly compatible platforms.

Let me know if you like this approach and will merge it and I will update the code accordingly.

LukasAud:
Hi @ghoneycutt, I'm sorry for the delay in feedback and/or reviewal. The puppet modules team (recently rebranded as CAT team) has been experiencing some major changes over the last few months. 

We have reviewed your PR again and discussed thoroughly whats the best approach to deal with the Solaris platform compatibility. We considered your and @binford2k opinion and finally decided to go ahead with approving and merging your PR.

LukasAud (approved):

ghoneycutt:
Thanks @LukasAud

alexjfisher:
I've updated the PR title to make it more explicit in the changelog that support (not that any of it necessarily worked anymore!) has been removed, not such deprecated.

PR #1419, opened 2021-07-08
Support Debian 9 and 10

ghoneycutt:


ghoneycutt:
Attempting to reproduce https://tickets.puppetlabs.com/browse/IAC-1595

PR #1420, opened 2021-07-08
Set ordering of acceptance tests

ghoneycutt:
This allows pre_run to be removed from the spec_helper_acceptance as the
tests will run in a specific order. The pre_run just included the
mysql::server class which now happens first thing.


ghoneycutt:
Based off of @pmcmaw 's previous work in trying to debug acceptance testing with Debian.

daianamezdrea:
Hi @ghoneycutt, every acceptance test should be independent, if we merge this PR, we'll create dependencies between the tests, which we don't want. Thank you ! (https://softwareengineering.stackexchange.com/questions/46913/writing-acceptance-test-cases)

ghoneycutt:
The acceptance tests already run in a specific order and when someone adds a new file that order will change. This only makes what is already happening explicit.

Also the tests are inherently dependent on each other because they are making changes to a system and checking the state of the system at that time. Unless you run each test block from the code on its own system in the same starting state, this will be the case and our testing frameworks (litmus and beaker) and processes in the community do not support that.

ghoneycutt:
@daianamezdrea @pmcmaw Could you please have a look at this.

adrianiurca (approved):

adrianiurca:
Hey @ghoneycutt, thank you for your contribution!

kind regards,
@adrianiurca

PR #1425, opened 2021-08-03
Solve issue with repeated restarts if ssl-disable is true

markasammut:
With the current code, if ssl-disable is set to true and restart is set to true, the notify message in the server::config class warning about ssl being disabled causes the service to be restarted on each Puppet run.

I have therefore moved that warning message outside of the server::config class, so that server::config ~> server::service does not trigger a restart of the service on each Puppet run.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -174,7 +174,7 @@ class mysql::server (
                 if you are forced to use a mysql version compiled without SSL support',
     }
   }
-  
+
   Anchor['mysql::server::start']
   -> Class['mysql::server::config']
   -> Class['mysql::server::install']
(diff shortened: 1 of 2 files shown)


markasammut:
PR will pass test If scientific7 and oraclelinux7 are excluded from xtrabackup tests as per https://github.com/puppetlabs/puppetlabs-mysql/pull/1443

daianamezdrea:
Hello @markasammut , sorry for the late answer but our team went through some transformations, can you rebase your PR in order to have everything green and ready to merge? Thank you and sorry again

markasammut:
Rebased, workflows awaiting approval.

LukasAud:
Reopening to trigger tests again.

LukasAud (approved):

PR #1427, opened 2021-08-05
MODULES-8373 Fix mysql_grant resource to be idempodent on MySQL 8+

theq86:
When the MySQL module is used against MySQL8, mysql_grant is no more idempodent.
This is because MySQL 8+ no more returns "ALL" for SHOW GRANTS, when all privileges
are granted. Instead, the separate privileges are shown.
Unfortunately, the tests were just adapted to not error in this situation.
This commit aims to fix the issue and also fixes the tests.

### DIFF
--- a/lib/puppet/provider/mysql_grant/mysql.rb
+++ b/lib/puppet/provider/mysql_grant/mysql.rb
@@ -46,8 +46,11 @@ Puppet::Type.type(:mysql_grant).provide(:mysql, parent: Puppet::Provider::Mysql)
           end
         end
         sorted_privileges = stripped_privileges.sort
-        if self.newer_than('mysql' => '8.0.0')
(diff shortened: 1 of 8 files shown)


david22swan:
@theq86 I'm seeing some rubocop failures on your changes, but their easy to fix.
All you need to do is install the bundle environment and then run `bundle exec rake rubocop -a`.
This will autocorrect any of the failures that it can and tell you what ones you'll need to fix manually.

Once that's done we can see about getting this merged.

theq86:
@david22swan Do you have a suggestion how to deal with "Style/GlobalVars: Do not introduce global variables." ?

The code that changes utf8 to utf8mb3 for MySQL 8 is needed in several places throughout the tests. To avoid code duplication I introduced the global charset variable in the spec_helper_acceptance.rb file.

I do not know ruby well enough to come up with a more elaborate way-

david22swan:
@theq86 Looking at this, firstly this code should not be in `spec_helper_acceptance.rb`, this file is managed entirely by the pdksync and should never be touched. Sorry for not catching it earlier.
Any changes such as this should instead go into the `spec_helper_acceptance_local.rb`, which is left untouched by the pdksync.

In regards to the code itself, wrapping it in a method call should work. Something like this:
```
def fetch_charset()
@charset ||= if os[:family] == 'ubuntu' && os[:release] =~ %r{^20\.04}
                        "utf8mb3"
                      else
                        "utf8"
                      end
end
```
This should set and return the value of `@charset` the first time the method is called and then just return it every other time after.

david22swan (approved):
@theq86 This looks good to me so I'm gonna go ahead and merge it :)
Thank you for getting back on my comments so fast :)

PR #1440, opened 2021-09-23
(IAC-1595) Return Debian to Metadata

david22swan:
Debian had previously been removed in order to cancel the tests, however this can now be accomplished with the `exclude_platforms` parameter within the sync.yml

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: '0 0 * * *'
 
+
 env:
   HONEYCOMB_WRITEKEY: 7f3c63a70eecc61d635917de46bea4e6
   HONEYCOMB_DATASET: litmus tests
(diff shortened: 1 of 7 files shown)


.github/workflows/nightly.yml
```diff
@@ -45,29 +46,27 @@ jobs:
...
         echo STEP_ID=Setup-Acceptance-Test-Matrix >> $GITHUB_ENV
         echo STEP_START=$(date +%s) >> $GITHUB_ENV
-
     - name: Setup Acceptance Test Matrix
       id: get-matrix
       if: ${{ github.repository_owner == 'puppetlabs' }}
       run: |
         if [ '${{ github.repository_owner }}' == 'puppetlabs' ]; then
-          buildevents cmd $TRACE_ID $STEP_ID matrix_from_metadata -- bundle exec matrix_from_metadata_v2
+          buildevents cmd $TRACE_ID $STEP_ID matrix_from_metadata -- bundle exec matrix_from_metadata_v2 --exclude-platforms '["Debian-9", "Debian-10", "Debian-11"]'
```
bastelfreak, line 61: why don't you want to run the tests on Debian?

david22swan: There is a problem with the percona backup feature that is causing consistent failures that require some changes that  we have not yet had the time to resolve. IT is on our backlog however and we will get to it when we have the time.

daianamezdrea (approved):

PR #1449, opened 2021-12-28
Use MariaDB for Ubuntu 20.04

treydock:
Ran into numerous issues when Puppet attempted to install MySQL 8.0. Not sure if maybe the repos on Azure Ubuntu 20.04 are the problem but using MariaDB solved the issues.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -171,9 +171,10 @@ class mysql::params {
     }
 
     'Debian': {
-      if $::operatingsystem == 'Debian' {
+      if $::operatingsystem == 'Debian' or
+      ($::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '20.04') >= 0) {
         $provider = 'mariadb'
-      } else { # Ubuntu
+      } else {
         $provider = 'mysql'
       }
       if $provider == 'mariadb' {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -171,8 +171,8 @@ class mysql::params {
     }
 
     'Debian': {
-      if $::operatingsystem == 'Debian' or
-      ($::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '20.04') >= 0) {
+      if $facts['os']['name'] == 'Debian' or
+      ($facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '20.04') >= 0) {
         $provider = 'mariadb'
       } else {
         $provider = 'mysql'
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -34,7 +34,7 @@ class mysql::server::config {
   }
 
   #Debian: Creating world readable directories before installing.
-  case $::osfamily {
+  case $facts['os']['family'] {
     'Debian': {
       if $managed_dirs {
         $managed_dirs.each | $entry | {
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -34,7 +34,7 @@ class mysql::server::config {
   }
 
   #Debian: Creating world readable directories before installing.
-  case $::operatingsystem {
+  case $::osfamily {
     'Debian': {
       if $managed_dirs {
         $managed_dirs.each | $entry | {
--- a/spec/acceptance/00_mysql_server_spec.rb
+++ b/spec/acceptance/00_mysql_server_spec.rb
@@ -38,8 +38,7 @@ describe 'mysql class' do
           },
           databases => {
             'somedb' => {
-              ensure  => 'present',
-              charset => #{fetch_charset},
+              ensure => 'present',
             },
           }
         }
--- a/spec/acceptance/01_mysql_db_spec.rb
+++ b/spec/acceptance/01_mysql_db_spec.rb
@@ -14,7 +14,6 @@ describe 'mysql::db define' do
         mysql::db { 'spec1':
           user            => 'root1',
           password        => 'password',
-          charset         => #{fetch_charset},
         }
       MANIFEST
     end
@@ -43,7 +42,6 @@ describe 'mysql::db define' do
           user     => 'root1',
           password => 'password',
           sql      => '/tmp/spec.sql',
-          charset  => #{fetch_charset},
         }
       MANIFEST
     end
@@ -68,7 +66,6 @@ describe 'mysql::db define' do
           user     => 'root1',
           password => 'password',
           dbname   => 'realdb',
-          charset  => #{fetch_charset},
         }
       MANIFEST
     end
--- a/spec/acceptance/04_mysql_backup_spec.rb
+++ b/spec/acceptance/04_mysql_backup_spec.rb
@@ -12,7 +12,6 @@ describe 'mysql::server::backup class' do
         ]:
           user     => 'backup',
           password => 'secret',
-          charset  => #{fetch_charset},
         }
 
         class { 'mysql::server::backup':
@@ -73,7 +72,6 @@ describe 'mysql::server::backup class' do
           ]:
             user     => 'backup',
             password => 'secret',
-            charset  => #{fetch_charset},
           }
 
           class { 'mysql::server::backup':
@@ -138,7 +136,6 @@ describe 'mysql::server::backup class' do
           ]:
             user     => 'backup',
             password => 'secret',
-            charset  => #{fetch_charset},
           }
           case $facts['os']['family'] {
             /Debian/: {
@@ -259,7 +256,6 @@ describe 'mysql::server::backup class' do
           ]:
             user     => 'backup',
             password => 'secret',
-            charset  => #{fetch_charset},
           }
           case $facts['os']['family'] {
             /Debian/: {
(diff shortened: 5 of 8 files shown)


ghoneycutt:
+1 tested that this is working

ghoneycutt:
@LukasAud This is another quick win that we have been using with a customer.

alexjfisher:
Seems reasonable. @treydock Would you consider this a breaking change, or is the module effectively broken for Ubuntu 20 users without this?
Given Ubuntu 22.04 came out in the last few days, should the PR title be updated?  (Or perhaps that should wait for a separate PR that adds ubuntu 22.04 to the test matrix and to metadata.json)

alexjfisher:
(and could someone restart the build so that we can see what the ubuntu 20.04 build failures were about?)

treydock:
> Seems reasonable. @treydock Would you consider this a breaking change, or is the module effectively broken for Ubuntu 20 users without this? Given Ubuntu 22.04 came out in the last few days, should the PR title be updated? (Or perhaps that should wait for a separate PR that adds ubuntu 22.04 to the test matrix and to metadata.json)

This module was not working at all on Ubuntu 20.04 without this change, at least not without overriding package names.  I think can wait till 22.04 can be tested.

manifests/params.pp
```diff
@@ -171,9 +171,10 @@
     }
 
     'Debian': {
-      if $::operatingsystem == 'Debian' {
+      if $::operatingsystem == 'Debian' or
```
alexjfisher, line 174: I appreciate a lot of the existing code doesn't, but can we start using the `$facts` hash here instead of using topscope variables?

eg.
```suggestion
      if $facts['operatingsystem'] == 'Debian' or
```

alexjfisher: (I've not any strong opinion on continuing to use 'legacy facts', but you could use the newer ones like `$facts['os']['name']` instead)

treydock: Updated to use non-legacy facts from facts hash.

LukasAud:
Hey @ghoneycutt, thanks for the ping, will be keeping a close eye on this one.

On the other hand, @treydock could you elaborate a bit more on the failure? This issue is very similar to another one we are experiencing on a cloud testing level for Debian systems. Any information and/or solutions for this issue will be very appreciated and could speed up work on our side for this and other issues.

treydock:
> On the other hand, @treydock could you elaborate a bit more on the failure? This issue is very similar to another one we are experiencing on a cloud testing level for Debian systems. Any information and/or solutions for this issue will be very appreciated and could speed up work on our side for this and other issues.

@LukasAud I no longer have access to the systems where this was an issue but it was Azure based Ubuntu 20.04 so this could be something unique to Azure's Ubuntu repos.  I also no longer have access to Azure based systems so not sure I'll be able to further test solutions and my normally deployed systems are all RHEL so can't test there either.

LukasAud:
@treydock That is okay, we are working on our side to troubleshoot a very similar issue on a different system. We will let you know if we find a reliable fix for the issue.

LukasAud:
Hi @treydock, there is a new release for our MySQL module which I think may affect your PR. Could you rebase it?

treydock:
I've rebased.

LukasAud:
Hi, @treydock. I have looked over the latest checks. The goods news is that the previous error, which would crash the test run, is no longer present. However, there are still two distinct failures in the acceptance run that need to be addressed before I can merge this PR. 

One seems to be a recurring problem over most of the logged failures. I have noticed this log which repeats over all failures and seems to point to the cause of the problem:
`Notice: /Stage[main]/Mysql::Server::Providers/Mysql_database[somedb]/charset: charset changed 'utf8' to 'utf8mb3'`

The other failure affects a single test case and throws the following errors:
```
Error: /Stage[main]/Main/Mysql::Db[spec2]/Exec[spec2-import]: Failed to call refresh: 'cat /tmp/spec.sql | mysql spec2' returned 1 instead of one of [0]
       Error: /Stage[main]/Main/Mysql::Db[spec2]/Exec[spec2-import]: 'cat /tmp/spec.sql | mysql spec2' returned 1 instead of one of [0]
```

If you can take a look at the code and address these errors so that the acceptance runs pass, I will be happy to merge this.

treydock:
> One seems to be a recurring problem over most of the logged failures. I have noticed this log which repeats over all failures and seems to point to the cause of the problem: `Notice: /Stage[main]/Mysql::Server::Providers/Mysql_database[somedb]/charset: charset changed 'utf8' to 'utf8mb3'`

If I had to guess, MariaDB for Ubuntu doesn't default to utf8mb3 so the changes to use that charset would need to be removed and just go back to utf8.  Does this sound acceptable?

> 
> The other failure affects a single test case and throws the following errors:
> 
> ```
> Error: /Stage[main]/Main/Mysql::Db[spec2]/Exec[spec2-import]: Failed to call refresh: 'cat /tmp/spec.sql | mysql spec2' returned 1 instead of one of [0]
>        Error: /Stage[main]/Main/Mysql::Db[spec2]/Exec[spec2-import]: 'cat /tmp/spec.sql | mysql spec2' returned 1 instead of one of [0]
> ```

Further down this is the actual error:

```
       Notice: /Stage[main]/Main/Mysql::Db[spec2]/Exec[spec2-import]/returns: ERROR 1050 (42S01) at line 1: Table 'table1' already exists
```

I think what's happening is the idempotent issue from charset is causing the test to refresh and try and import the SQL again.  Let me know of changing the default charset to utf8 for Ubuntu 20.04 is acceptable and I'll make that change.

LukasAud:
Yes, as long as the change is necessary and it doesn't break any other functionality, we are happy to accept this sort of changes.

Furthermore, it seems like the only instance of charset replacement occurs in a spec_helper (spec_helper_acceptance_local.rb) file, and thus, any change or code removal should only affect the test cases, and no other core module functionality.

LukasAud (approved):

LukasAud:
That's perfect. Thanks for your contribution.

daehnomel:
I've found that Several errors will occur when trying to apply ANY grants to the root@localhost MySQL user in a manifest using the forge MySQL module Version 11.03 and higher, this has happened on 6 hosts in my environment with the configuration below.  Removing all grants to root@localhost allows MySQL AND or MariaDB to install AND OR continue operating, without error.
Puppet Community Version 7.20
Ubuntu 20.04
MySQL 8.0.31

daehnomel:
> I've found that Several errors will occur when trying to apply ANY grants to the root@localhost MySQL user in a manifest using the forge MySQL module Version 11.03 and higher, this has happened on 6 hosts in my environment with the configuration below. Removing all grants to root@localhost allows MySQL AND or MariaDB to install AND OR continue operating, without error. Puppet Community Version 7.20 Ubuntu 20.04 MySQL 8.0.31

Additionally, abandoning MySQL for MariaDB on Ubuntu, as done here,  breaks prior implementations of this manifest by attempting to install MariaDB over MySQL.  Explicitly declaring "package name' and 'package_ensure' resource types, as recommended does not properly override the MariaDB install issue.  These issues cause significant problems for production systems running Ubuntu / MySQL and updating this module from previous versions, especially in concert with the issue listed above.  Please reconsider making MariaDB the default, or, make overriding that an easier affair.

enkidushane:
The change to MariaDB as the default in 20.04 caused issues for me as well.  I started installing mysql-server and mysql-client with a version of puppetlabs-mysql that predated the release of Ubuntu 20.04, and while the server installed as expected, the client failed.  So I updated the module to the latest release, and couldn't understand why it was complaining about mariadb-server not being able to install.

After looking at the change log, I figured I'd go ahead and remove mysql-server and let it install mariadb, and while it installed the mariadb server and client packages, it would not start (complaining about "/usr/sbin/mysqld: One can only use the --user switch if running as root".  I spent 30 minutes trying to understand that error, finally noticed that the "user" setting was coming from puppet, and decided I needed to purge maridb and mysql, totally wipe out /etc/mysql and re-run puppet.  After doing that, mariadb server still wouldn't start, but for some other reason.

So this time I purged / deleted /etc/mysql, and then "downgraded" to v12.0.3 of puppetlab-mysql.  Re-ran puppet and mysql server and client installed without issue, mysql server started fine, and puppet set up the 100+ users / databases as configured.   I don't know what issue were originally encountered with Mysql 8.0 on Ubuntu 20.04, but running a pretty generic setup, I didn't encounter any major issues (there's a minor issue involving default charset and collate, but that was easy to address in my local module).

A comment in another ticket (https://github.com/puppetlabs/puppetlabs-mysql/issues/1500#issuecomment-1302474078) possibly suggests (the wording is ambiguous enough that I'm not sure) that MariaDB is what gets installed on Ubuntu 20.04 when you install mysql-server, but I don't think that's the case.

I know the switch to mariadb as the default in puppetlabs-mysql is nearly a year old now, so reverting this change may itself be breaking, but perhaps one or more of the following could be done:

LukasAud:
Hi @daehnomel and @enkidushane, we always appreciate feedback. However, its hard for us to track community opinions, issues and suggestions when they are in dead/merged PRs. If you have an issue to report or a suggestion to improve the module, please feel free to open a discussion in our '[Issues](https://github.com/puppetlabs/puppetlabs-mysql/issues)' tab and, if relevant, link back to this PR for more information. That way, we will have better visibility on it.

PR #1450, opened 2022-02-08
pdksync - (IAC-1787) Remove Support for CentOS 6

david22swan:
(IAC-1787) Remove Support for CentOS 6
pdk version: `2.2.0`


chelnak (approved):
✅

Issue #1455, opened 2022-03-18
charset and collate must both be set if specifying a non default charset for a database


benjamin-robertson:
Currently, charset and collate parameters for the defined type mysql::db don't mention in the docs they need to both set to the same charset. EG
```
  mysql::db { 'bensdb':
    user     => ben,
    password => lookup('mysql::server::root_password'),
    host     => 'localhost',
    grant    => 'ALL',
    charset  => 'latin1',
    collate  => 'latin1_swedish_ci',
  }
```

If they are not both set; corrective changes continually occur. 
```
Notice: /Stage[main]/Profile::Lab::Mysql/Mysql::Db[bensdb]/Mysql_database[bensdb]/collate: collate changed 'latin1_swedish_ci' to 'utf8_general_ci' (corrective)
Next run
Notice: /Stage[main]/Profile::Lab::Mysql/Mysql::Db[bensdb]/Mysql_database[bensdb]/charset: charset changed 'utf8mb3' to 'latin1' (corrective)
Next run
Notice: /Stage[main]/Profile::Lab::Mysql/Mysql::Db[bensdb]/Mysql_database[bensdb]/collate: collate changed 'latin1_swedish_ci' to 'utf8_general_ci' (corrective)
```

I propose changing the parameter descriptions from

```
# @param charset
#   The character set for the database.
# @param collate
#   The collation for the database.
```

To 
```
# @param charset
#   The character set for the database. Must be set with collate to avoid corrective changes.
# @param collate
#   The collation for the database. Must be set with charset to avoid corrective changes.
```

PR #1456, opened 2022-03-18
Updated db defined type strings documentation (#1)

benjamin-robertson:
Updates to Puppet strings documentation for mysql module.

Regenerated Reference.md

Fixes issue #1455

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1138,7 +1138,7 @@ Default value: `$name`
 
 Data type: `Any`
 
-The character set for the database.
+The character set for the database. Must be set with collate to avoid corrective changes.
 
 Default value: `'utf8'`
 
@@ -1146,7 +1146,7 @@ Default value: `'utf8'`
 
 Data type: `Any`
 
-The collation for the database.
+The collation for the database. Must be set with charset to avoid corrective changes.
 
 Default value: `'utf8_general_ci'`
 
@@ -1556,7 +1556,7 @@ Hash a string as mysql's "PASSWORD()" function would do it
 
 #### `mysql::password(Variant[String, Sensitive[String]] $password, Optional[Boolean] $sensitive)`
 
-Hash a string as mysql's "PASSWORD()" function would do it
+The mysql::password function.
 
 Returns: `Variant[String, Sensitive[String]]` hash
 The mysql password hash from the clear text password.
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -18,9 +18,9 @@
 # @param dbname
 #   The name of the database to create.
 # @param charset
-#   The character set for the database.
+#   The character set for the database. Must be set with collate to avoid corrective changes.
 # @param collate
-#   The collation for the database.
+#   The collation for the database. Must be set with charset to avoid corrective changes.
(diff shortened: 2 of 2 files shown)


chelnak:
Hey! Thanks for your contribution. Would you be able to provide a little more context around your proposed change?

Thanks 😊

benjamin-robertson:
Hi Craig,

It appears that both the charset and collate parameters both need to be set. If only charset is set we continually receive corrective changes on Puppet runs. 

For example if you configure the following
```
  mysql::db { 'bensdb':
    user     => ben,
    password => lookup('mysql::server::root_password'),
    host     => 'localhost',
    grant    => 'ALL',
    charset  => 'latin1',
  }
```
Corrective changes will continually occur for collate.

See notes from below issue ticket.

Thanks

Ben



Currently, charset and collate parameters for the defined type mysql::db don't mention in the docs they need to both set to the same charset. EG
```
  mysql::db { 'bensdb':
    user     => ben,
    password => lookup('mysql::server::root_password'),
    host     => 'localhost',
    grant    => 'ALL',
    charset  => 'latin1',
    collate  => 'latin1_swedish_ci',
  }
```
If they are not both set; corrective changes continually occur.
```
Notice: /Stage[main]/Profile::Lab::Mysql/Mysql::Db[bensdb]/Mysql_database[bensdb]/collate: collate changed 'latin1_swedish_ci' to 'utf8_general_ci' (corrective)
Next run
Notice: /Stage[main]/Profile::Lab::Mysql/Mysql::Db[bensdb]/Mysql_database[bensdb]/charset: charset changed 'utf8mb3' to 'latin1' (corrective)
Next run
Notice: /Stage[main]/Profile::Lab::Mysql/Mysql::Db[bensdb]/Mysql_database[bensdb]/collate: collate changed 'latin1_swedish_ci' to 'utf8_general_ci' (corrective)
```
I propose changing the parameter descriptions from
```
# @param charset
#   The character set for the database.
# @param collate
#   The collation for the database.
```
To

```
# @param charset
#   The character set for the database. Must be set with collate to avoid corrective changes.
# @param collate
#   The collation for the database. Must be set with charset to avoid corrective changes.
```

chelnak:
Hey Ben, thanks for such a detailed response.

I think what you are suggesting makes sense and I also think that we could probably do some more in that area to make the defined type a bit more robust. 

That would take some time right now so I think a docs update is a good interim solution.

How about we make the descriptions read something like:

"Must have the same value as X to avoid corrective changes"

I think that makes the intention a little more explicit.

What do you think?

benjamin-robertson:
Hi Craig,

Updated as suggested; I also provided a link to the mysql page which explains how to view the charset and collate pairs. Since the names don't match exactly. 

Thoughts?

Ben

chelnak:
This is great.

There is a double space at the start of "Must" on line 49. If you could fix that so it matches the previous description then we can get this merged!

chelnak (changes requested):
Just one small change then we are looking good!

benjamin-robertson:
Space fixed. Sneaky space.

chelnak (approved):

PR #1466, opened 2022-04-12
(Bugfix) Grant privileges idempotency Fix

LukasAud:
Prior to this commit, the Ubuntu 20.04 spec testing was failing two
test cases at mysql_grant_spec. One of them was a simple mismatch
between the expected stdout and actual stdout of the test case, and the
other had to do with a misbehaviour from the module at the time of
retrieving privileges from users. In the last MySQL version, the module, 
upon retrieving privileges from an user, creates and array that contains 
both 'ALL' and all dynamic privileges which is later used to check for 
idempotency in certain test cases, causing an error in idempotency.

This commit aims to the previously mentioned errors. The first error is
a simple 'expected message' adjust whereas for the second issue, a
temporary workaround that removes the additional dynamic privileges 
so that testing does not fail.

In the future, the issue should be assessed and a fix that actually
modifies the behaviour of the module to account for such processing of 
privileges should be implemented.


david22swan (approved):
LGTM

PR #1479, opened 2022-07-12
Add CentOS 9 to supported operating systems

kajinamit:
This adds CentOS 9 to supported operating systems.


david22swan:
@kajinamit Unfortunately I'm gonna have to close this.
Support for CentOS 9 is still waiting to be declared as supported by the Agent and so we must wait as well.
Given that Redhat 9 is supported however CentOS should work for you out of the box, we simply can't declare it.
Apologies

PR #1481, opened 2022-07-20
[Compatibility] Add Raspbian OS to provider configuration

jordi-upc:
On Raspbian OS must be defined as provider 'mariadb', otherwise can not install mysql-agent

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -170,7 +170,7 @@ class mysql::params {
     }
 
     'Debian': {
-      if $facts['os']['name'] == 'Debian' or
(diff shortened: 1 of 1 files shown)


LukasAud:
Hi @jordi-upc, this is change that we are happy to merge. However, we want to point out that this is only a compatibility change as we do not currently support Raspbian. I will make a small adjustment to the PR title before merging just to avoid any confusion about that topic.

Thanks for your contribution! 👍

LukasAud (approved):
SLES-15 test failures are unrelated.

PR #1482, opened 2022-08-02
Facter: confine to linux nodes

buchey8070:
Having this module loaded on windows nodes, will raise an error like:
`error while resolving custom facts in C:/ProgramData\PuppetLabs\puppet\cache\lib\facter\mysql_server_id.rb: C:/ProgramData/PuppetLabs/puppet/cache/lib/facter/mysql_server_id.rb:20: syntax error, unexpected keyword_rescue, expecting keyword_end
rescue
^
C:/ProgramData/PuppetLabs/puppet/cache/lib/facter/mysql_server_id.rb:23: syntax error, unexpected keyword_end, expecting end-of-input`

Confine Facter to only linux kernels prevents this error.


lib/facter/mysql_server_id.rb
```diff
@@ -14,9 +14,12 @@ def mysql_id_get
 end
 
 Facter.add('mysql_server_id') do
+  confine kernel: :linux
```
bastelfreak, line 17: I'm not sure if this is supposed to work on FreeBSD. In that case you could confine to not Windows

alexjfisher: As per my comment [here](https://github.com/voxpupuli/puppet-prometheus/pull/630#issuecomment-1202506554), confining to Linux has nothing to do with the issue you're encountering (very old ruby on your Windows agents still try to load facts even if they don't run them).

mobalt: Actually this is a ruby 2.5 feature as explained in https://stackoverflow.com/a/51726301/11953415 

This bug started showing up on our CentOS7 systems after rubocop to introduce the incompatible syntax in an April commit. https://github.com/puppetlabs/puppetlabs-mysql/commit/8095cce0affc3b8359a129685cef2f036dd1ae87

This autocommit by rubocop was in response to this change in March 2022 in which rubocop targets ruby version 2.5 instead of 2.4 https://github.com/puppetlabs/puppetlabs-mysql/commit/4ebc431250d0d665dbaca0824be2377caa2c3479#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5c 

Could we just target ruby 2.4 again as per version 12.0.2?

bastelfreak: Why Ruby 2.4? Which Puppet Agent/Ruby Version do you use? Puppet packages Ruby 2.5/Puppet 6 and Ruby2.7/Puppet 7 in their AIO packages.

LukasAud:
Closing and reopening PR to re-kick the tests.

LukasAud:
Hi @buchey8070, thanks for your interest in contributing to the module. There are a couple of issue that need resolving before we can decide whether this can be merged or not. 

First, there seem to be a few failing test cases for SLES-12 OS. Our nightlies show that these test are passing our daily checks so it is likely that the issue is related to the changes. Also, our Unit tests (Spec Tests / foobar) seem to be crashing because of a Rubocop offence. That one should be easy to fix.

Finally, some community members started a review that may have some insights on the problem you are facing. Most of our active trusted contributors have a fair amount of experience and their advice and/or alternative solutions tend to be useful in improving the PRs.

lib/facter/mysql_server_id.rb
alexjfisher: @mobalt Thanks for the correction.  I must have been testing something different when I decided it was a ruby 2.0 thing.  The not having anything to do with confining to Linux is still the case though.

Targeting ruby 2.4 vs 2.5 is probably best discussed in the pdk-templates repo.  Specifically see this PR. https://github.com/puppetlabs/pdk-templates/pull/466

LukasAud:
Hi @buchey8070, it seems like your PR has been stale for a while. We are currently cleaning up our modules and trying to address any old PRs. Since there hasn't been any engagement since the last few comments, we will be closing down this one. However, be aware that closed PRs can still be reopened if the author wants to resume work on them.

Best of luck!

PR #1484, opened 2022-08-16
Harden db defined type

chelnak:
Prior to this commit there was a possibility that malformed strings could be passed in to the resource. 
This could lead to unsafe executions on a remote system.

The parameters that are susceptible are `dbname`, `sql` and `import_cat_cmd`.

In addition, commands were not properly broken out in to arrays of arguments when passed to the exec resource.

This commit fixes the above by adding validation to parameters (where possible) to ensure that the given values conform to expectation.

`dbname` is validated with a regular expression that ensures that it matches expectations set by mysql.

`sql` now only accepts values as an `Array[String]`. 
This can be one or more value. 
Values are also validated with a regular expression that ensures that the given paths are valid.

`import_cat_cmd` has been removed in favour of a boolean parameter called `use_zcat`.
Setting this parameter to `true` will set the value of the private `import_cat_cmd` to `zcat`.

This commit also contains some minor refactoring.

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -9,6 +9,12 @@
 #     grant    => ['SELECT', 'UPDATE'],
 #   }
 #
+# @param name
+#   The name of the database to create. Database names must:
+#     * be longer than 64 characters.
+#     * not contain / \ or . characters.
+#     * not contain characters that are not permitted in file names.
+#     * not end with space characters.
 # @param user
 #   The user for the database you're creating.
 # @param password
@@ -28,7 +34,7 @@
 # @param grant_options
 #   The grant_options for the grant for user@host on the database.
 # @param sql
-#   The path to the sqlfile you want to execute. This can be single file specified as string, or it can be an array of strings.
+#   The path to the sqlfile you want to execute. This can be a an array containing one or more file paths.
 # @param enforce_sql
 #   Specifies whether executing the sqlfiles should happen on every run. If set to false, sqlfiles only run once.
 # @param ensure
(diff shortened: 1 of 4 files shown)


LivingInSyn (approved):

LivingInSyn (approved):

LivingInSyn (dismissed):

binford2k:
I don't think that `use_zcat` is sufficient. For example, one might want to use `bzcat`. What if we kept `import_cat_command` but specified its data type as `Enum['cat', 'zcat', 'bzcat']`? Are there other commands that make sense? I don't think we need to support `bat` and all the other fancy pants alternatives, just posix.

One could argue for the ability to specify a full path, like `/usr/local/bin/cat` but I think that's a bad idea and opens us up to undefined behavior.

LivingInSyn:
> I don't think that `use_zcat` is sufficient. For example, one might want to use `bzcat`. What if we kept `import_cat_command` but specified its data type as `Enum['cat', 'zcat', 'bzcat']`? Are there other commands that make sense? I don't think we need to support `bat` and all the other fancy pants alternatives, just posix.
> 
> One could argue for the ability to specify a full path, like `/usr/local/bin/cat` but I think that's a bad idea and opens us up to undefined behavior.

I think we should have it as an enum with the most common options in there. We can always accept PRs for new ones

chelnak:
Enum sounds sensible to me... and actually removes a breaking element of the change.

jelinwils (dismissed):

manifests/db.pp
```diff
@@ -41,25 +47,41 @@
...
+  $charset                                      = 'utf8',
+  $collate                                      = 'utf8_general_ci',
+  $host                                         = 'localhost',
+  $grant                                        = 'ALL',
+  $grant_options                                = undef,
+  Optional[Array] $sql                          = undef,
+  $enforce_sql                                  = false,
+  Enum['absent', 'present'] $ensure             = 'present',
+  $import_timeout                               = 300,
+  Enum['cat', 'zcat'] $import_cat_cmd           = 'cat',
```
binford2k, line 61:
```suggestion
  Enum['cat', 'zcat', 'bzcat'] $import_cat_cmd  = 'cat',
```

binford2k (dismissed):
I do think that we need bzcat in the command options, otherwise this looks good

GSPatton (approved):
all LGTM

Rathios:
This has already been merged, but .. you can easily support both string and array in the `sql` parameter by doing `flatten([$sql])`. Would have saved some needless breakage.

```Error while evaluating a Resource Statement, Mysql::Db[powerdns]: parameter 'sql' expects a value of type Undef or Array, got String (file: /etc/puppetlabs/code/environments/dev/modules/powerdns/manifests/backends/mysql.pp, line: 70) on node x```

apoleon:
Hello,

is this the fix for CVE-2022-3276 or are #1487 #1486 1485 also needed to resolve it ?

chelnak:
You will need all of them. We recommend consuming the lates major release.

PR #1485, opened 2022-08-17
Harden root password class

chelnak:
Prior to this PR there was a possibility that malformed strings could be passed in to the resource. 
This could lead to unsafe executions on a remote system.

The parameters that are susceptible are `install_secret_file`.

This PR fixes the above by adding validation to ensure the given values confirm to expectation.

`secret_file` is validated with a regular expression that ensures the given value is a valid path.


chelnak:
ping @puppetlabs/security

chelnak:
The test failures in this PR are present on main and part of another investigation.

manifests/server/root_password.pp
```diff
@@ -19,6 +19,12 @@
   $secret_file = $mysql::server::install_secret_file
   $login_file = $mysql::server::login_file
 
+  # Validate that the given secret_file value is a valid path
+  if $secret_file !~ /^(?:\/[A-Za-z0-9._-]+)+?$/ {
```
LivingInSyn, line 23: I think this regex needs some work, we might actually want to parse this with a ruby function if it's exposed instead of trying to regex it. Maybe something like:

```
secret_file = Shellwords.escape($secret_file)
if(File.file?($secret_file))
... 
```

chelnak: Agree that the regex is a bit shady..

shell_escape is exposed via stdlib so we can use that to clean the input.

I did want some sort of validation  happen though.. unless that what File.file? would provide in your example?

manifests/server/root_password.pp
```diff
@@ -16,26 +16,19 @@
...
   # New installations of MySQL will configure a default random password for the root user
   # with an expiration. No actions can be performed until this password is changed. The
   # below exec will remove this default password. If the user has supplied a root
   # password it will be set further down with the mysql_user resource.
   $rm_pass_cmd = join([
-      "mysqladmin -u root --password=\$(grep -o '[^ ]\\+\$' ${secret_file}) password ''",
-      "rm -f ${secret_file}",
+      "mysqladmin -u root --password=\$(grep -o '[^ ]\\+\$' /.mysql_secret) password ''",
+      'rm -f /.mysql_secret',
   ], ' && ')
```
LivingInSyn, line 28: This should be chained with a `;` not `&&`. If the `mysqladmin` command fails with the current command chaining it won't delete the `.mysql_secret` file

binford2k: I think that's desired though. If we remove the file without setting the password, then we've lost it.

LivingInSyn: 🤔 we don't want to leave a cleartext file sitting on the hard drive though

manifests/server/root_password.pp
binford2k: We cannot do actual filesystem validation because this runs during compile only. The best we can do is string validation that it's a valid pathname.

One thing we could do for this class of problem is to adopt my [assert resource](https://github.com/binford2k/puppet-assert) into stdlib and conditionally skip resources if assertions (like File.file?) were to fail. That's probably a different conversation though.

manifests/server/root_password.pp
binford2k: Let's look at the workflow and fragilities. This is how I understand it, please correct me if I'm wrong.

1. the mysql install runs and generates a randomized password
2. this is saved in plaintext on disk. It allows you only the ability to set your own password--you can't use it for normal operations.
3. the user is expected to specify a new password and delete the temporary password file, which is what this command string is doing.

If step 3 fails, I can only think of a handful of causes:
1. mysql installation failed or the service didn't start. Manual intervention is required, and the plaintext password may or may not be valid.
2. race condition -- the install succeeded, but an unknown actor already changed the root password. The plaintext password is invalid, but this should probably trigger an investigation.
3. the process that generated the install password and file on disk failed. The plaintext password is invalid.

Are there other scenarios?

manifests/server/root_password.pp
chelnak: This code has been removed now, so I'm resolving this.

manifests/server/root_password.pp
LivingInSyn: yes, assume that something else went wrong with the `mysqladmin` password change command, in this case the password was not changed or removed, and it's still sitting on the disk, right? The safer option here is to just delete it regardless of success or failure

binford2k: I'd still call that scenario 1 and the proper response imho is to fail and ensure manual intervention. I'm not sure if I care about the password file existing. It could make troubleshooting easier, but I suspect that most users won't bother--they'll either burn the node or remove the installation and try again.

My concern with chaining with `;` is that it will make that exec always succeed and so it won't bubble up errors.

LivingInSyn: how about we do it like this

```shell
 mysqladmin -u root --password=\$(grep -o '[^ ]\\+\$' /.mysql_secret) password \
&& (rm -f  /.mysql_secret; exit 0) \
|| (rm -f /.mysql_secret; exit 1)
 ```
 Now we'll always delete the secret, and we'll have the error if something went wrong?

binford2k: Ha! I looked at the original code again and it wasn't surfacing errors to start with!

LivingInSyn (dismissed):

LivingInSyn (approved):

LukasAud (approved):

PR #1487, opened 2022-08-19
Harden config class

chelnak:
Prior to this commit the variables `dir` and `mysql::server::package_name` were passed to `exec` resources in such a way that could allow unsafe executions on the remote host.

This commit fixes the above by properly parameterizing the arguments passed to each `exec` resource.

Additionally the variables been sanitized with `shell_escape` for good measure.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -39,14 +39,19 @@ class mysql::server::config {
       if $managed_dirs {
         $managed_dirs.each | $entry | {
           $dir = $options['mysqld']["${entry}"]
+
           if ( $dir and $dir != '/usr' and $dir != '/tmp' ) {
+            $clean_dir = shell_escape($dir)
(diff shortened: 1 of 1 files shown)


LivingInSyn (approved):

binford2k (dismissed):

david22swan (approved):

kajinamit:
I don't know whether it is feasible to document the requirement in metaata.json but this change is dependent on the change in puppet which is available since 7.9.0/6.24.0 and breaks compatibility with older versions.

chelnak:
Hey, this is a really good observation.. the answer is hard too given that we currently support such a wide range of versions.

Technically users are not required to update to the latest and greatest version of a module but this obviously means they miss out on newer features or fixes.

@binford2k What are your thoughts?

PR #1488, opened 2022-08-23
(MAINT) Fix acceptance tests

chelnak:
This PR contains various fixes for acceptance tests.

## Debian / Ubuntu

* Breaking due to the mysqld apparmor profile present on the GitHub Runner
* Fixed by removing profile prior to docker provisions

## RHEL 9

* Breaking on xtrabackup tests
* Fixed by instructing rspec not to run xtrabackup tests when running on RHEL 9

## SLES 15

* Breaking due to a charset mismatch
* Fixed by ensuring that the correct charset is configured during the test setup

This PR also:

* Restores the use of `matrix_from_metadata_v2` in workflows
* Removes the `export_locales` method along with references and tests because it is no longer needed.


GSPatton (commented):
code looks good. Will see how the tests do before approving

GSPatton (approved):
Approving since all tests passed (for the first time in a while)

LukasAud (approved):

Issue #1499, opened 2022-10-10
Path validation of sql parameter in mysql::db too restrictive.


rwf14f:
## Describe the Bug

The regex that is used by `mysql::db` to validate the `sql` parameter refuses to accept perfectly valid pathnames, eg it only accepts a single dot to separate the filename from the extensions, but not anywhere else in the path. In our case it's the powerdns mysql schema path that it's choking on: _/usr/share/doc/pdns-backend-mysql-4.1.14/schema.mysql.sql_.

Is there a reason why the module doesn't use the types from stdlib (`Optional[Array[Stdlib::Unixpath]] $sql = undef,`)? The module has a dependency on it already. Are you worried about whitespaces in paths ?

## Expected Behavior

The module should allow all valid paths.

## Steps to Reproduce

Use paths with a multiple dots in sql parameter.

## Environment
 - Version [13.0.0]
 - Platform [CentOS 7]

LukasAud:
This issue has been addressed in this PR https://github.com/puppetlabs/puppetlabs-mysql/pull/1505

@rwf14f Can you confirm if this a valid solution for the issue?

rwf14f:
Yes, it is. The updated regex validates my use cases correctly now.

TwizzyDizzy:
Hi folks,

TBH, I think that the elephant in the room hasn't been properly answered:

> Is there a reason why the module doesn't use the types from stdlib (Optional[Array[Stdlib::Unixpath]] $sql = undef,)? The module has a dependency on it already. Are you worried about whitespaces in paths ?

Cheers
Thomas

PR #1505, opened 2022-10-25
mysql::db sql parameter support filenames with multiple dots

skn-bvdh:
This fixes mysql::db sql parameter to support filenames with multiple dots. #1499

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -76,8 +76,8 @@ define mysql::db (
   # Ensure that the sql files passed are valid file paths.
   if $sql {
     $sql.each | $sqlfile | {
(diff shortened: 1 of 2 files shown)


leemachine:
The change in this PR still does not support absolute filenames that have more than one digit on the directories name.

For example, `/opt/app/tmp/app-1.2.3/schema/app.sql`

LukasAud:
Hi @skn-bvdh, your change looks good and it seems like most of our tests are passing fine. However, there seems to be an indentation error that is triggering our Rubocop checks and stopping our Spec tests from running properly. 

If you fix that small issue and the tests pass, we should be happy enough to merge this.

LukasAud (approved):

LukasAud:
LGTM. Thanks for your contribution!

PR #1511, opened 2022-11-08
1509: Fix ssl params issue for EL7 and newer mariadb versions

loopiv:
#1509 Fix ssl params issue with newer versions of mariadb on el7

Note: previous PR had cla assistant problem that I could not resolve

PR #1512, opened 2022-11-24
1509: fix ssl for debian Mariadb

elfranne:
#1511 is a fix for EL7, same fix for Debian

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -200,9 +200,9 @@ class mysql::params {
       $mycnf_owner             = undef
       $mycnf_group             = undef
       $socket                  = '/var/run/mysqld/mysqld.sock'
-      $ssl_ca                  = '/etc/mysql/cacert.pem'
-      $ssl_cert                = '/etc/mysql/server-cert.pem'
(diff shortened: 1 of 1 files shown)


markruys:
A workaround is to add `ssl-disable: true` to your config. Although then you must live with:
```
Notice: /Stage[main]/Mysql::Server/Notify[ssl-disable]/message: current_value 'absent', should be "Disabling SSL is evil! You should never ever do this except\n
                if you are forced to use a mysql version compiled without SSL support" (noop)
```
Seriously? Module maintainers, please remove this silly repeating message. We're not kids, so let us be responsible for our own acts :)

chelnak:
@markruys Hi! We would welcome a PR if you'd like to change the presentation of this message.

Alternatively you can open a new issue and ask for its removal.

Thanks,
Craig

LukasAud:
Hi @elfranne, this issue has already been addressed in the following [PR](https://github.com/puppetlabs/puppetlabs-mysql/pull/1513). If the issue still persists for you despite this update, please let us know. Otherwise, this PR can be closed.

Issue #1509, opened 2022-11-08
SSL cert/key params in server.cnf break new versions of mariadb.  SSL is disabled.


loopiv:
## Describe the Bug
With ssl set to false, the module still adds the cert, ca, and key params which break the latest version of mariadb on EL7.

## Expected Behavior
Those params should not be present when ssl is disabled.

## Steps to Reproduce
1. On EL7, upgrade mariadb to latest version as of 08-Nov-2022.
2. MariaDB fails to start.

## Environment
 - Version = 12.0.1
 - Platform = CentOS 7

## Additional Context
We're not enabling ssl so it's set to false as default.

vanElden:
This is not only an issue on RedHat systems; it's just the same on Ubuntu.

TwizzyDizzy:
I agree, that this is a broader issue and needs a broader (in terms of operating systems) fix than the one suggested in https://github.com/puppetlabs/puppetlabs-mysql/pull/1511.

That being said, I think it calls for a definiate decision on whether TLS should be enabled by default or not. Given the changes in MariaDB, it seems a decision to enable TLS by default would entail to also generate proper certs/keys by this module.

Since generating the keys properly has problems in itself ("self-signed" is only marginally better than unencrypted traffic) the go-to solution should probably be to disable TLS by default (so set those parameters to `undef` - but every operating systems in contrast to the PR only changing this for RedHat).

All that being said, one might argue that this is "only a MariaDB issue" since they now introduced this change to abort the daemon startup, but I think this assessment is wrong: the TLS context hasn't been initialised properly on other operating systems, too. The module just didn't care because it had no consequences but from an abstract point of view, this is broken because it does not provide secure encryption by default.

Cheers
Thomas

**Side note** (and I'm happy to keep this out of the scope of this issue): given that this module plays [TLS police](https://github.com/puppetlabs/puppetlabs-mysql/blob/v13.0.1/manifests/server.pp#L167-L172) in other places I would say that it requires quite a level of resistance to cognitive dissonance, that you would warn in one case, but happily configure invalid encryption in another place just for the sake of "having enabled encryption" (assuming you have certs/keys in the default locations).

vmpr:
this issue broke my database after an update to the latest version `10.5.18-MariaDB` on Debian.
the parameters need to be `undef` otherwise every puppet run will break our system again.
is there a quick way to fix that until the module is fixed? i can't change the code my self as we pull direct via r10k from puppet forge.

[...]

PR #1513, opened 2022-11-25
do not emit other ssl directives when ssl = false

kjetilho:
This is needed for newer versions of MariaDB, see issue #1509


LukasAud:
Hi, @kjetilho. Thanks for your contribution. We are currently investigating the issue you are trying to address. Unfortunately, it doesnt seem like our test cases cover this scenario so we will need to investigate a bit further to ensure that your PR does not have any unexpected side effect. Meanwhile, could you implement a test case for your change?

kjetilho:
something like the above?

LukasAud:
Yes, that should be enough. Thanks for your contribution. We will merge as soon as we finish testing it on our side.

LukasAud (approved):
LGTM

Issue #1516, opened 2022-12-14
Documentation out of date for mysql::db sql parameter


kreczko:
## Describe the Change You Would Like

Following current examples and documentation leads to an error:
```bash
Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Mysql::Db[apel]: parameter 'sql' expects a value of type Undef or Array, got String
```
&rarr; documentation related to the sql parameter is incorrect.

### manifests/db.pp
```puppet
# @param sql
#   The path to the sqlfile you want to execute. This can be an array containing one or more file paths.
```
&rarr; needs to be an array of one or more files (or undef) - not sure about correct phrasing.

### README.md
```puppet
mysql::db { 'mydb':
  ...
  sql             => '/path/to/sqlfile.gz',
  ...
}
```
needs to be
```puppet
mysql::db { 'mydb':
  ...
  sql             => ['/path/to/sqlfile.gz'],
  ...
}
```

pmcmaw:
Hey @kreczko thank you for highlighting this issue, this has been an oversight on our end and I will work on resolving it. Our code is dynamically documented in the REFERENCE.md therefore is usually more up to date. I have created a PR to address this issue, again thank you for taking the time to raise this issue #1517

kreczko:
@pmcmaw np, happy to help. After all, this is a very useful module :).

> Our code is dynamically documented in the REFERENCE.md therefore is usually more up to date

Missed that one. I assume it is taking the info directly from the module docs since it references the same sentence: "The path to the sqlfile you want to execute.".
While the type is mentioned, the above sentence can easily be misinterpreted (`path != array`).

pmcmaw:
Fix has now been merged :-) Im going to close this issue.
If you are waiting on a release just let me know. 👍

PR #1517, opened 2022-12-14
(GH-1516) Update sql example to use array

pmcmaw:
Prior to this commit the db example shows a String however it is an optional array. 

This commit updates the example from a String to an array.


GSPatton (approved):
looks good

Issue #1518, opened 2022-12-19
exec failure for puppet < 6.24


trefzer:
## Describe the Bug
for puppet < 6.24 this module fails with error:
`Error: Failed to apply catalog: Parameter command failed on Exec[wait_for_mysql_socket_to_open]: Command must be a String, got value of class Array (file: /etc/puppetlabs/code/environments/dev_test/modules/mysql/manifests/server/service.pp, line: 57)`

## Expected Behavior
not failing !

## Steps to Reproduce
puppet below 6.24

## Analysis
The following commits added exec parameters as an Array which is not supported before [puppet 6.24 (release notes)](https://puppet.com/docs/puppet/6/release_notes_puppet.html#unique_1428117924)
* cdaa8393e3fe9b8f971945ee8cb1f3e933902a0d
* 1c1291d27883fc41eab666eb65f3c85071d6c696
* 90168d93834f329ca037f83bedc5b5e580955fb6

## Solution:
Either revert the usage of arrays in exec or raise puppet version needed (client version !) in metadata.json

pmcmaw:
Thank you for logging this issue @trefzer, we have resolved this issue and cut a release.

PR #1519, opened 2022-12-20
(GH-1518) Declare minimum Puppet version 6.24.0

pmcmaw:
In codebase hardening efforts the commands are passed as an array, but this feature was only introduced in Puppet 6.24.01. This raises the minimum version to match, since it's no longer possible to use the module on anything older.


GSPatton (approved):
looks good

PR #1521, opened 2022-12-20
Use consistent values for ssl-ca, ssl-cert and ssl-key

### LINKED ISSUE #1509: SSL cert/key params in server.cnf break new versions of mariadb.  SSL is disabled.

smortex:
The default value of the server's ssl-ca, ssl-cert and ssl-key
parameters are system-dependent:
  * Some systems set these parameters to `undef`;
  * Some systems set these parameters to a filename which is not managed
    by the module.

Recent versions of MariaDB insist on being able to read the files these
parameters point to even if TLS is not in use.  As a consequence,
MariaDB is broken and cannot start successfully.

Always use `undef` as the default version of these parameters so that
users have the same experience regardless of the Operating System they
are running.

Existing PR #1511 and #1512 propose a partial fix focused on the
current breakage of MariaDB on RedHat and Debian OS families.  This PR
is proposed as an alternative to bring consistent configuration across
all systems.

Fixes #1509

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -106,9 +106,6 @@ class mysql::params {
       $mycnf_owner             = undef
       $mycnf_group             = undef
       $socket                  = '/var/lib/mysql/mysql.sock'
-      $ssl_ca                  = '/etc/mysql/cacert.pem'
-      $ssl_cert                = '/etc/mysql/server-cert.pem'
-      $ssl_key                 = '/etc/mysql/server-key.pem'
       $tmpdir                  = '/tmp'
       $managed_dirs            = undef
       # mysql::bindings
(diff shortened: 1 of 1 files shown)


LukasAud:
Hi @smortex, thanks for your contribution. I'm taking a look around your PR and it looks good so far but I just wanted to make sure that it is still relevant. Last week we merged the following [fix](https://github.com/puppetlabs/puppetlabs-mysql/pull/1513/files) for issue #1509. Did this fix not address properly the issue?

smortex:
@LukasAud good catch! I missed the release that fixed the root cause and was following up on #1511 and #1512.  I guess these PR can be closed too.

LukasAud:
No worries, thanks for taking the time to work on a solution. We always appreciate community contributions. 👍

PR #1522, opened 2022-12-20
xtrabackup.sh only touch when backup_success_file_path is set

JvGinkel:
By default `backup_success_file_path` is `undef` what result in a `touch` command without argument that will produce an error like `touch: missing file operand`.

With this merge request we only add the `touch` command to the script when the value of `backup_success_file_path` is defined. 

Also added rspec test for both cases.


LukasAud (approved):
LGTM

PR #1524, opened 2022-12-22
fix naming in documentation

jonasbebe:
This module configures no potgres, so we should describe the params with mysql instead of postgres.


LukasAud (approved):
LGTM

PR #1531, opened 2023-01-24
Allow to set and run backups using another user via cron 

deepak7340:


david22swan:
Functionality looks good on a first pass but would need some test's to back it up before we can merge it in.

LukasAud:
Hi @deepak7340, as @david22swan previously mentioned, the code looks good but we would need you to create some tests    to back it up for the future. Additionally, it would be great if you could create a better description for the PR.

It also looks like there are some conflicts that must be resolved before CI can start running.

If you can address those few blockers, we would be happy to review/merge this.

bronachfalls:
Hi @deepak7340 , due to inactivity I am going to close this PR. If at any point you would like for it to be reopened, please let me know and I'll happily do so.

PR #1532, opened 2023-01-25
(CONT-359) Syntax update

LukasAud:
Code now compliant with rules regarding:

legacy facts
parameter datatypes
parameter documentation
top-scope facts


manifests/params.pp
```diff
@@ -340,7 +340,7 @@
       $config_file         = '/etc/my.cnf'
       $includedir          = undef
       $datadir             = '/var/mysql'
-      $log_error           = "/var/mysql/${::hostname}.err"
+      $log_error           = "/var/mysql/${facts['os']['hostname']}.err"
```
smortex, line 343:
```suggestion
      $log_error           = "/var/mysql/${facts['networking']['hostname']}.err"
```

smortex (commented):
:wave: it seems like I had a few more legacy facts removed in `/spec/` in #1535

manifests/params.pp
LukasAud: Investigated this and confirmed a number of facts hidden in the test files that were not caught by our Lint plugin. Will amend soon.

LukasAud: That should be those addressed

smortex (approved):
Looks good!

david22swan (approved):
LGTM

jdyb:
Was it intentional to change `Mysql::Server` parameter `package_ensure` to no longer allow `latest` ?
It used to work for us and when we updated the module we hit the type restriction.

PR #1535, opened 2023-02-15
Drop legacy facts

smortex:
Puppet 8 will default to not send legacy facts. Update the module so that it uses the facts that superseded these legacy facts.


smortex:
RedHat 7 failure seems to be a transient one. Also the mend failure is caused by a lack of license key, so unrelated to this change.

This is ready for review!

smortex:
Close in favor of #1532

PR #1536, opened 2023-02-18
Also set passwords for the additional root users.

klynton:
MySQL/MariaDB also have root@127.0.0.1 and root@::1 users. If the password is not set for them then any user can log in as root without a password by specifying `-h 127.0.0.1` or `-h ::1`.

### DIFF
--- a/manifests/server/root_password.pp
+++ b/manifests/server/root_password.pp
@@ -35,6 +35,16 @@ class mysql::server::root_password {
       password_hash => Deferred('mysql::password', [$mysql::server::root_password]),
       require       => Exec['remove install pass'],
     }
+    mysql_user { 'root@127.0.0.1':
+      ensure        => present,
+      password_hash => mysql::password($mysql::server::root_password),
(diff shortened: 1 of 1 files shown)


LukasAud:
Hi @klynton, can you rebase this PR? That should kick off a run of our latest CI and test against Puppet 8. Once that is done, I will be happy to review this.

klynton:
@LukasAud Done! Thank you.

LukasAud:
Hey @klynton, looks like there are some failures. It looks like your code counts as a duplicate of another declaration existing in manifests/server/account_security.pp L7. Can you take a look into this?

klynton:
It looks like the module tries to delete those accounts now:

https://github.com/puppetlabs/puppetlabs-mysql/blob/main/manifests/server/account_security.pp#L7-L14

I'll have a look into this a bit more. But I think that removing those users is the same as setting the password for them in terms of security...unless I'm missing something.

Issue #1539, opened 2023-03-02
Mysql::Db parameter 'tls_options' expects a value of type Undef or Array, got String


alexbugl:
## Describe the Bug
After update of v13.1.0 to v13.2.0 of the _puppetlabs-mysql_ module we get the following error:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Mysql::Db[icinga2]:
parameter 'tls_options' expects a value of type Undef or Array, got String (file: /etc/puppetlabs/code/environments/dev/modules/icinga/manifests/database.pp, line: 94) on node X
```

## Expected Behavior
In version v13.2.0 you added the type `Optional[Array[String[1]]]` to the parameter `tls_options`, which breaks one of the  modules we use (_puppet-icinga_) -- this tries to create a MySQL DB using this code:
```puppet
    $_tls_options = if $tls {
      if $tls =~ String and $tls == 'cert' {
        'X509'
      } else {
        'SSL'
      }
    } else {
      'NONE'
    }

    mysql::db { $db_name:
      tls_options => $_tls_options,
      # shortened for readability
    }
```
So it writes a string into attribute `tls_options` of type `mysql::db`, which does not match the associated type.
But your code in `lib/puppet/type/mysql_user.rb` seems to still be able to cope with strings, as it creates an array out of a single string:
```ruby
  newproperty(:tls_options, array_matching: :all) do
    desc 'Options to that set the TLS-related REQUIRE attributes for the user.'
    validate do |value|
      value = [value] unless value.is_a?(Array)
      if value.include?('NONE') || value.include?('SSL') || value.include?('X509')
```

## Environment
 - Module Version v13.2.0
 - Platform Debian 11

## Suggested Solution
A possible solution could be to correct the type (untested at the moment):
```diff
-Optional[Array[String[1]]]                     $tls_options     = undef,
+Optional[Variant[String[1], Array[String[1]]]] $tls_options     = undef,
```

janit42:
To maybe speed a bugfix up a bit, I've created a pull request https://github.com/puppetlabs/puppetlabs-mysql/pull/1540 based on Alex suggestion.

alexbugl:
I tested the fix, and it really solved the bug. :)
Please apply ...

janit42:
[...]

PR #1540, opened 2023-03-03
allow tls_options to be of String

janit42:
the type of $tls_options has recently been changed to Array in 5131e0f806 while the code in lib/puppet/type/mysql_user.rb is able to work with it as an Array or a String.

This change allows $tls_options to by a String in addition to Array, which re-establishes compatibility with other modules, like puppet-icinga

Fixes https://github.com/puppetlabs/puppetlabs-mysql/issues/1539

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -49,7 +49,7 @@
 define mysql::db (
   String[1]                                      $user,
   Variant[String, Sensitive[String]]             $password,
-  Optional[Array[String[1]]]                     $tls_options     = undef,
+  Optional[Variant[String[1], Array[String[1]]]] $tls_options     = undef,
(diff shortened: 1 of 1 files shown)


bastelfreak:
@janit42 please rebase against main to get rid of the merge commit.

janit42:
> @janit42 please rebase against main to get rid of the merge commit.

thanks for the heads-up @bastelfreak ! Anything else I can do for now?

bastelfreak (approved):

bastelfreak:
@janit42 can you maybe add a unit test for it?

janit42:
I don't think this is needed anymore - compatibility has been ensured in related icingia module.

Issue #1502, opened 2022-10-21
Privileges not working correctly even after fix of issue #1474


o4ier:
## Describe the Bug
puppet makes following changes each run:
Notice: /Stage[main]/Mysql::Server::Providers/Mysql_grant[mydba@localhost/*.*]/privileges: privileges changed ['ALL', 'APPLICATION_PASSWORD_ADMIN', 'AUDIT_ABORT_EXEMPT', 'AUDIT_ADMIN', 'AUTHENTICATION_POLICY_ADMIN', 'BACKUP_ADMIN', 'BINLOG_ADMIN', 'BINLOG_ENCRYPTION_ADMIN', 'CLONE_ADMIN', 'CONNECTION_ADMIN', 'ENCRYPTION_KEY_ADMIN', 'FIREWALL_EXEMPT', 'FLUSH_OPTIMIZER_COSTS', 'FLUSH_STATUS', 'FLUSH_TABLES', 'FLUSH_USER_RESOURCES', 'GROUP_REPLICATION_ADMIN', 'GROUP_REPLICATION_STREAM', 'INNODB_REDO_LOG_ARCHIVE', 'INNODB_REDO_LOG_ENABLE', 'PASSWORDLESS_USER_ADMIN', 'PERSIST_RO_VARIABLES_ADMIN', 'REPLICATION_APPLIER', 'REPLICATION_SLAVE_ADMIN', 'RESOURCE_GROUP_ADMIN', 'RESOURCE_GROUP_USER', 'ROLE_ADMIN', 'SENSITIVE_VARIABLES_OBSERVER', 'SERVICE_CONNECTION_ADMIN', 'SESSION_VARIABLES_ADMIN', 'SET_USER_ID', 'SHOW_ROUTINE', 'SYSTEM_USER', 'SYSTEM_VARIABLES_ADMIN', 'TABLE_ENCRYPTION_ADMIN', 'XA_RECOVER_ADMIN'] to ['ALL'] (corrective)

## Expected Behavior
puppet should not change privileges that are already "ALL" to "ALL" in each run.

## Steps to Reproduce
1. Setup mysql-server with:
```
  class { '::mysql::server':
[...]
    users => {
      'mydba@localhost' => {
        ensure        => 'present',
        password_hash => <myhash>,
      },
    grants => {
      'mydba@localhost/*.*' => {
        ensure     => 'present',
        options    => ['GRANT'],
        privileges => ['ALL'],
[...]

PR #1542, opened 2023-03-15
add another workaround for mysql_grant provider for mysql >= 8.0.27

david-barbion:
Starting MySQL 8.0.27, a new privilege is added: FIREWALL_EXEMPT. When granting ALL PRIVILEGES, this new privilege is also granted.

This PR adds a new workaround for this use case.

Ref: https://dev.mysql.com/doc/refman/8.0/en/privileges-provided.html#priv_firewall-exempt


david-barbion:
This may also address issue #1502

PR #1544, opened 2023-03-17
allow metacharacters * and ? in sql file path

zivis:
Hi,

I had a problem with another module (sensson/puppet-powerdns)
which has given `/usr/share/doc/pdns-backend-mysql-*.*.*/schema.mysql.sql` as sql to mysql::db
this failed with `file … is invalid. A valid file path is expected.`

Are the metacharacters ? and * problematic in these paths and should I fix it else where or would it be okay to use these?

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -78,7 +78,7 @@ define mysql::db (
   # Ensure that the sql files passed are valid file paths.
   if $sql {
     $sql.each | $sqlfile | {
-      if $sqlfile !~ /^\/(?:.[.A-Za-z0-9_-]+\/?+)+(?:\.[.A-Za-z0-9]+)+$/ {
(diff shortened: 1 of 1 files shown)


BuJo:
Since https://github.com/sensson/puppet-powerdns/commit/26ca5b655544d12007cdd57f5c764f8751c47163 this might not be necessary anymore.

However, it says it depends on the puppet mysql 15 (judging by the changes this might not be truly necessary), which in turn needs stdlib 9 which is backwards incompatible rendering many third/first-party modules unuseable.  The new `powerdns` module might work, rendering the current PR moot?

bastelfreak (approved):

bronachfalls:
Hi @zivis , could you please rebase? Thank you!

bronachfalls:
Closing and reopening this PR to rerun fresh tests

PR #1548, opened 2023-03-24
pdksync - (CONT-768) Pin puppetlabs-puppet_agent

GSPatton:
(CONT-768) Pin puppetlabs-puppet_agent
pdk version: `2.5.0`


LukasAud (approved):

bastelfreak:
can you explain the change? Why is main unsafe? What kind of changes could it have that break puppetlabs-mysql? And if there are such changes, could they be a bug that needs to be catched and not ignored?

PR #1549, opened 2023-03-28
fixtures.yml: Delete legacy symlink

bastelfreak:
puppetlabs_spec_helper will create the symlink for us, we don't need it in .fixtures.yml.


alexjfisher (approved):
Yep.  Has done for _years_

alexjfisher (approved):

david22swan (approved):
LGTM

PR #1552, opened 2023-03-28
move static data from params to server class

bastelfreak:
This enables puppet-strings to pick up the data.


david22swan (approved):
LGTM

Imo, params should only contain variables used in multiple manifests, or ones that vary depending on the OS or other facts

PR #1564, opened 2023-05-10
Fix broken sensitive parameter for mysql::password

### LINKED ISSUE #1563: Sensitive parameter support broken of mysql::password
## Describe the Bug
The puppet function `mysql::password` does not allow a second parameter anymore to toggle the return value to be a sensitive. Instead the puppet run fails with a wrong number of arguments message.

This issue has been introduced with #1557
https://github.com/puppetlabs/puppetlabs-mysql/commit/effc6846e132fe0c41d279cb1ff03e989d4e3cb7#diff-f63c8340e97cb15646da94767358abc703a41ba144fa217d9585e00b8a04b289L22-R22

## Expected Behavior
- Use
  ```
  mysql('password', true)
  ```
- Returns
  ```
  sensitive('*2470C0C06DEE42FD1618BB99005ADCA2EC9D1E19')
  ```

## Steps to Reproduce
- Use
  ```
  mysql('password', true)
  ```
- Fails with
  ```
  Error while evaluating a Function Call:
  
  wrong number of arguments (given 2, expected 1)
  ```

## Environment
 - Module v14.0.0
 - Puppet 7.24.0
 - OracleLinux 7

## Additional Context
Note: That the rspec test for this case is currently simply skipped instead of testing it. This needs to also be fixed via the same PR.
https://github.com/puppetlabs/puppetlabs-mysql/blob/00255c75fa8cd7cf112facf77c844535fe6a71fc/spec/functions/mysql_password_spec.rb#L26-L30

cruelsmith:
Revert broken change of #1557 and fixes the spec test for it by removing the skip.

Fixes #1563

### DIFF
--- a/lib/puppet/functions/mysql/password.rb
+++ b/lib/puppet/functions/mysql/password.rb
@@ -19,7 +19,7 @@ Puppet::Functions.create_function(:'mysql::password') do
     return_type 'Variant[String, Sensitive[String]]'
   end
 
-  def password(password, sensitive: false)
+  def password(password, sensitive = false)
     password = password.unwrap if password.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
 
     result_string = if %r{\*[A-F0-9]{40}$}.match?(password)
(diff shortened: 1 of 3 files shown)


b4ldr (dismissed):

cruelsmith:
Sry @b4ldr could you check it please once again? I missed the RuboCop ignore rule since i just run the spec test locally.

b4ldr (approved):

jordanbreen28 (approved):
LGTM. Thank you @cruelsmith

PR #1569, opened 2023-05-19
(CONT-576) allow deferred function for token & secrets

Ramesh7:
- implementing deferred function for password field


spec/unit/puppet/provider/mysql_login_path/mysql_login_path_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
```
Ramesh7, line 3: Observed failure on my local with uninitialized 
```
  2) Puppet::Provider::MysqlLoginPath::MysqlLoginPath create(context, name, should) creates the resource
     Failure/Error: let(:context) { instance_double(Puppet::ResourceApi::BaseContext) }

     NameError:
       uninitialized constant Puppet::ResourceApi::BaseContext
```

manifests/backup/mysqlbackup.pp
```diff
@@ -108,14 +108,14 @@
...
-      'password'               => $backuppassword_unsensitive,
+      'password'               => Deferred('mysql::password', [$backuppassword_unsensitive]),
     },
   }
   $options = mysql::normalise_and_deepmerge($default_options, $mysql::server::override_options)
 
   file { 'mysqlbackup-config-file':
     path    => '/etc/mysql/conf.d/meb.cnf',
-    content => template('mysql/meb.cnf.erb'),
+    content => deferrable_epp('mysql/meb.cnf.erb', $options),
```
alexjfisher, line 118:
```suggestion
    content => stdlib::deferrable_epp('mysql/meb.cnf.erb', $options),
```
??
and you'd need to bump the minimum stdlib version in metadata.json too.

Ramesh7: @alexjfisher thanks for your comment, btw I am searching for one of working example with `deferrable_epp` which I haven't come across yet. If you have any can you please share?
The doc is also bit old (seems updated when Puppet 7 added with this feature)  https://www.puppet.com/docs/puppet/7/template_with_deferred_values.html

Ramesh7: Looks like my stdlib wasn't installed on local, that was leading to function not defined. After installing it worked fine. Please ignore above comment. Thanks

Ramesh7: Have updated supported version as well into metadata file as well based on changelog of stdlib : https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/CHANGELOG.md#v840-2022-07-21

PR #1593, opened 2023-08-16
params.pp: cleanup service provider logic

bastelfreak:
the service provider will figure it out on its own. We don't need to duplicate the logic from the providers into puppet code.


kajinamit (approved):
Makes very clear sense

Ramesh7 (approved):

Issue #1576, opened 2023-06-15
xtrabackup


ubellavance:
## Describe the Bug
Error when running puppet with an xtrabackup config created

## Expected Behavior
`puppet apply` should run w/o error

## Steps to Reproduce
1. Manifest to get configuration from hiera:
```
  # Create backup configuration in Hiera
  if $mysql_backup_config {
    # Using create_resources
    # create_resources('class', $mysql_backup_config, {})
    # or
    # using each(preferable)
    $mysql_backup_config.each |$backup_class,$backup_config| {
      class { $backup_class:
        * => $backup_config,
      }
    }
  }

```
2. Hiera configuration :
```
profile::mariadb::mysql_backup_config:
  'mysql::backup::xtrabackup':
    backupuser: 'pbackupuser'
    backuppassword: 'mypassword'
    backupmethod: 'mariabackup'
    backupmethod_package: 'mariadb-backup'
    backupdir: '/var/backups/mysql/ddnfs/dev/physical'
    backuprotate: 7
    execpath: '/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin'
    time: [ '23' , '15' ]
```

## Environment
 - Puppet 7.25
 - RHEL8

PR #1595, opened 2023-08-29
Ensure fact mysql_version exist when granting privileges to xtrabacku…

ndelic0:
…p user in MySQL 8.0

## Summary
PR should allow setting up xtrabackup configuration without breaking initial puppet run as mysql_version fact is not available in the catalog.

## Additional Context

## Related Issues (if any)
PR should fixed the issue reported https://github.com/puppetlabs/puppetlabs-mysql/issues/1576

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
Reproduced issue
[root@localhost production]# puppet apply -e 'include profile::mariadb'
```
Warning: Module 'profile': Value returned from data_hash function 'yaml_data', when using location '/etc/puppetlabs/code/environments/production/modules/profile/data/common.yaml', must use keys qualified with the name of the module
Error: Evaluation Error: Error while evaluating a Function Call, 'versioncmp' parameter 'a' expects a String value, got Undef (file: /etc/puppetlabs/code/environments/production/modules/mysql/manifests/backup/xtrabackup.pp, line: 53, column: 8) on node localhost.localdomain
```
with fix applied 

```
Notice: Compiled catalog for localhost.localdomain in environment production in 0.31 seconds
Notice: /Stage[main]/Profile::Mariadb/Package[mysql]/ensure: created
Notice: /Stage[main]/Profile::Mariadb/Package[mariadb]/ensure: created
Notice: /Stage[main]/Mysql::Server::Config/File[/etc/my.cnf.d]/ensure: created
Notice: /Stage[main]/Mysql::Server::Config/File[mysql-config-file]/ensure: defined content as '{sha256}e44307efdb246c98f0320b6f34b3cfc3931b25e3b6fa2f337b6b3ace65825567'
Notice: /Stage[main]/Mysql::Server::Install/Package[mysql-server]/ensure: created
```
in next puppet run backup configs will be applied.


jst-cyr:
This PR seems to have been neglected for review and testing, but given the limited changes in here and the lack of activity I am going to close this PR for now. If this type of a change is needed in the future, we can re-open.

PR #1597, opened 2023-09-20
(CAT-1428) Removal of redhat/scientific/oraclelinux 6 for mysql module

praj1001:
## Summary
With the removal of support for RHEL6 from our modules, work must be done to cleanup and remove any residual code.
This PR does the same for mysql repo module. More info can be found #1498 

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -485,7 +485,7 @@ class mysql::params {
   }
 
   ## Additional graceful failures
-  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '4' and $facts['os']['name'] != 'Amazon' {
-    fail("Unsupported platform: puppetlabs-${module_name} only supports RedHat 6.0 and beyond.")
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] < '7' and $facts['os']['name'] != 'Amazon' {
+    fail("Unsupported platform: puppetlabs-${module_name} only supports RedHat 7.0 and beyond.")
   }
 }
--- a/spec/acceptance/05_mysql_xtrabackup_spec.rb
+++ b/spec/acceptance/05_mysql_xtrabackup_spec.rb
@@ -40,7 +40,7 @@ describe 'mysql::server::backup class with xtrabackup', if: Gem::Version.new(mys
           /RedHat/: {
             # RHEL/CentOS 5 is no longer supported by Percona, but older versions
             # of the repository are still available.
-            if versioncmp($facts['os']['release']['major'], '6') >= 0 {
+            if versioncmp($facts['os']['release']['major'], '7') >= 0 {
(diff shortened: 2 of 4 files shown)


spec/acceptance/05_mysql_xtrabackup_spec.rb
```diff
@@ -40,7 +40,7 @@ class { 'mysql::server': root_password => 'password' }
           /RedHat/: {
             # RHEL/CentOS 5 is no longer supported by Percona, but older versions
             # of the repository are still available.
-            if versioncmp($facts['os']['release']['major'], '6') >= 0 {
```
Ramesh7, line 43: if versioncmp($facts['os']['release']['major'], '7') >= 0 {

praj1001: Sure Thanks

manifests/params.pp
```diff
@@ -486,6 +486,6 @@
 
   ## Additional graceful failures
   if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '4' and $facts['os']['name'] != 'Amazon' {
```
david22swan, line 488: This should be updated to check for RedHat less than 7

spec/classes/mysql_backup_xtrabackup_spec.rb
```diff
@@ -42,8 +42,6 @@ class { 'mysql::server': }
                       'percona-xtrabackup-24'
                     elsif Puppet::Util::Package.versioncmp(facts[:os]['release']['major'], '7') >= 0
```
david22swan, line 43: This should just be an else now

spec/classes/mysql_backup_xtrabackup_spec.rb
```diff
@@ -197,8 +195,6 @@ class { 'mysql::server': }
...
```
david22swan, line 196: same as above.

david22swan (changes requested):
Few small things

spec/classes/mysql_backup_xtrabackup_spec.rb
david22swan: My thought is that an if/elsif should always end in an else, so you have the if OS is 8 or above  followed by the else that catches 7.

manifests/params.pp
praj1001: Missed this one. Thanks for pointing out @david22swan

spec/classes/mysql_backup_xtrabackup_spec.rb
praj1001: Yeah, my bad. I completely agree. I misunderstood that but its clear now. Thanks for pointing out @david22swan

david22swan (approved):

Ramesh7 (approved):

PR #1610, opened 2023-12-11
Fix backup/rotation with multiple excluded databases

BuJo:
## Summary

Right now, excluding multiple databases does not work.
This MR aims to fix excluding multiple databases, by revising the respective regular expression.

## Additional Context

* When using multiple excluded databases, the list of databases is filtered using `grep -v`. i.e. `grep -v '^\(information_schema|performance_schema\)$`
* When using Basic vs Extended Regular Expressions, the characters `(` and `|` lose their special meaning, the backslashed versions have to be used. For the group (`()`) the escaping has been done, however the alternation is unescaped.

Leading to:

* All the excluded databases will be backed up.
* In case a database is not backuppable (which is why it had been excluded), this leads to the cleanup not being run at all, as it depends on the backup having been successful.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/spec/classes/mysql_backup_mysqldump_spec.rb
+++ b/spec/classes/mysql_backup_mysqldump_spec.rb
@@ -76,13 +76,13 @@ describe 'mysql::backup::mysqldump' do
         let(:params) do
           {
             'file_per_database' => true,
-            'excludedatabases' => ['information_schema']
+            'excludedatabases' => ['information_schema', 'performance_schema']
           }.merge(default_params)
         end
 
         it {
           expect(subject).to contain_file('mysqlbackup.sh').with_content(
-            %r{information_schema},
+            %r{information_schema\\\|performance_schema},
           )
         }
       end
--- a/templates/mysqlbackup.sh.epp
+++ b/templates/mysqlbackup.sh.epp
@@ -84,7 +84,7 @@ cleanup
 <% if $excludedatabases.empty { -%>
 mysql --defaults-extra-file=$TMPFILE -s -r -N -e 'SHOW DATABASES' | while read dbname
 <%} else {-%>
-mysql --defaults-extra-file=$TMPFILE -s -r -N -e 'SHOW DATABASES' | grep -v '^\(<%= $excludedatabases.join('|') %>\)$' | while read dbname
+mysql --defaults-extra-file=$TMPFILE -s -r -N -e 'SHOW DATABASES' | grep -v '^\(<%= $excludedatabases.join('\\|') %>\)$' | while read dbname
 <% } -%>
 do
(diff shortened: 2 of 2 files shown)


BuJo:
Branch is working in production without problems.

BuJo:
Should I rebase this branch on top of the current `main`?

spec/classes/mysql_backup_mysqldump_spec.rb
```diff
@@ -76,13 +76,13 @@ class { 'mysql::server': }
...
             'file_per_database' => true,
-            'excludedatabases' => ['information_schema']
+            'excludedatabases' => ['information_schema', 'performance_schema']
           }.merge(default_params)
         end
 
         it {
           expect(subject).to contain_file('mysqlbackup.sh').with_content(
-            %r{information_schema},
+            %r{information_schema\\\|performance_schema},
```
bastelfreak, line 85: Aren't that two \ too much?

```
root@webserver01 ~ # mysql -s -r -N -e 'SHOW DATABASES' | grep -v '^\(bastelfreak_blog\\|bastelfreak_past\)$'
bastelfreak_blog
bastelfreak_past
bastelfreak_wp
information_schema
mysql
nextcloud
performance_schema
sys
root@webserver01 ~ # mysql -s -r -N -e 'SHOW DATABASES' | grep -v '^\(bastelfreak_blog\|bastelfreak_past\)$'
bastelfreak_wp
information_schema
mysql
nextcloud
performance_schema
sys
root@webserver01 ~ # mysql -s -r -N -e 'SHOW DATABASES' | grep -v '^\(bastelfreak_blog\\\|bastelfreak_past\)$'
bastelfreak_blog
bastelfreak_wp
information_schema
mysql
nextcloud
performance_schema
sys
root@webserver01 ~ #
```

Or is that just for the escaping due to %r being a regular expression?

bastelfreak:
I was wondering why no tests catched the issue you described. Turns out there's a test, but only for ancient mysql: https://github.com/puppetlabs/puppetlabs-mysql/pull/1642.

spec/classes/mysql_backup_mysqldump_spec.rb
BuJo: Yeah, the three escapes are necessesary because we are matching a regular expression with a regular expression.  The `|` and `\` having special meanings on all "layers".

BuJo:
Should I try and port the test forward before merging @bastelfreak ?

BuJo:
@bastelfreak I rebased the branch on top of current `main`, hopefully the tests go through.

PR #1612, opened 2023-12-28
Add ability to use hex hash with caching_sha2_password plugin

C24-AK:
# Summary
This is a pull request to add the ability for users to use hex hashes to authenticate with caching_sha2_password plugin.

## How does it work:

```
mysql::server::users:
  native@localhost:
    password_hash: '*8A47DD31CC0202C1682AA9C78AAD8D031E817BCB'  # this is mysql_native_password
  caching@localhost:
    plugin: caching_sha2_password
    password_hash: '0x244124303035244A306D441C066316643B317F7C361706720F2D6A4C45772F7050666A7032656B7443544A5543683556386A654A3367427336484E4A2E3361666C6868454743' # this is a hex hash for caching_sha2_password
```
   
As for now, you can't use plain passwords, since the module does not create a valid hash for you.
But it is possible to use a generated hex hash.

## Why not the original authentication string?
The original authentication string when printed out contains non-printable binary characters.
Since the hex format always contains a unique string with printable characters, this is working.

## How to generate this hash?
Locally I used this to generate my hash and then use it in puppet:

```
CREATE USER 'caching'@'localhost' IDENTIFIED WITH 'caching_sha2_password' BY '<your-password>';
SET print_identified_with_as_hex=1; 
SHOW CREATE USER 'caching'@'localhost'; 
DROP USER 'caching'@'localhost'; 
SET print_identified_with_as_hex=0;
```

This gives you the output. It is necessary to have the 0x prefix in front of the hash, so mysql syntax will not be broken.

I appreciate every commit and help for this PR. I am by no mean a ruby expert and used it for the first time.


Jimadine:
This would be a very useful improvement ❤

To add to this, looking at the defined type `mysql::db`, it appears that there is no user `plugin` parameter as there is for `mysql_user`.  `mysql::db` uses `mysql_user`, so I don't think there's any reason why it can't.  So an improvement to your PR would be to add a user plugin parameter to `mysql::db`.

I've quickly modified `manifests/db.pp` with [what I think would need changing](https://gist.github.com/Jimadine/5673997a0140ce9dbf54035bceffda45) (Lines 8,23,24,55 & 110 have been added). For comparison, [here's the original](https://github.com/puppetlabs/puppetlabs-mysql/blob/38fad7503639f4f1d526ca6313e6bc2d357046ed/manifests/db.pp) from `main`.

I'm not sure if the new `plugin` parameter I've added should be an enum, or whether it should be a bit more relaxed as it is in `mysql_user`, which is a regex matching `%r{\w+}`.

Note that, in `mysql::db`, the existing `password` parameter value can be either a password or password hash, as it uses `mysql_user`.

C24-AK:
Hey @Jimadine ,
I appreciate the answer!
Ye I think that's a good idea. I would prefer the plugin to be more relaxed, since we don't know every plugin (maybe there are custom plugins?), regex matching is fine.

I encountered a bug which I will focus first and come back to the request later. :)

CiderAndWhisky:
Can we get this merged? Would be helpful for us, we use this branch in prod already, and we want to be able to update without losing the changes.

bastelfreak:
@C24-AK thanks for the PR! Can you take a look at the rubocop annotations and rebase against the latest main branch (not a merge commit)?

C24-AK:
@bastelfreak done! Looking forward for the review

Jimadine:
@C24-AK - Do you think you'd be able to add a `plugin` parameter to `$user_resource` of the `mysql::db` defined type, per my [earlier comment](https://github.com/puppetlabs/puppetlabs-mysql/pull/1612#issuecomment-1904503968), as part of this PR, or would you prefer to keep this separate?

C24-AK:
> @C24-AK - Do you think you'd be able to add a `plugin` parameter to `$user_resource` of the `mysql::db` defined type, per my [earlier comment](https://github.com/puppetlabs/puppetlabs-mysql/pull/1612#issuecomment-1904503968), as part of this PR, or would you prefer to keep this separate?

Done, I appreciate if you could test the changes and see if everything works as expected

manifests/db.pp
```diff
@@ -47,21 +50,22 @@
...
+  String[1]                                              $collate         = 'utf8_general_ci',
+  String[1]                                              $host            = 'localhost',
+  Variant[String[1], Array[String[1]]]                   $grant           = 'ALL',
+  Optional[Variant[String[1], Array[String[1]]]]         $grant_options   = undef,
+  Optional[Array]                                        $sql             = undef,
+  Boolean                                                $enforce_sql     = false,
+  Enum['absent', 'present']                              $ensure          = 'present',
+  Integer                                                $import_timeout  = 300,
+  Enum['cat', 'zcat', 'bzcat']                           $import_cat_cmd  = 'cat',
+  Optional[String]                                       $mysql_exec_path = undef,
```
bastelfreak, line 68: Personally I'm not a fan of reindenting all parameters because that makes the commit hard to sherry-pick/rebase/revert (in case we ever need that). However I don't know which kind of styleguide Puppet prefers here.

C24-AK: Reverted it

Jimadine:
> > @C24-AK - Do you think you'd be able to add a `plugin` parameter to `$user_resource` of the `mysql::db` defined type, per my [earlier comment](https://github.com/puppetlabs/puppetlabs-mysql/pull/1612#issuecomment-1904503968), as part of this PR, or would you prefer to keep this separate?
> 
> Done, I appreciate if you could test the changes and see if everything works as expected

Thank you! I've looked at your changes and they look good. I'm attempting to test, but am facing a problem related to setting the mysql root password:
```
Error: can't modify frozen String: "ALTER USER 'root'@'localhost' IDENTIFIED WITH"
Error: /Stage[main]/Mysql::Server::Root_password/Mysql_user[root@localhost]/password_hash: change from [old password hash redacted] to [new password hash redacted] failed: can't modify frozen String: "ALTER USER 'root'@'localhost' IDENTIFIED WITH"
```
I'm not sure if this problem relates to the changes in this PR, but I tried `15.0.0` from Puppet Forge and didn't see the problem. The Puppet resource relevant to the error is:
```
class {'::mysql::server':
  package_name            => 'mysql-server',
  service_name            => 'mysql',
  remove_default_accounts => true,
  restart                 => true,
  service_enabled         => true,
  config_file             => '/etc/mysql/my.cnf',
  create_root_user        => true,
  create_root_my_cnf      => true,
  root_password           => $dlib_atom::mysql_root_password,
  override_options        => {
    mysqld => {
      default-authentication-plugin => $mysql_user_auth_plugin,
      sql_mode                      => 'ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION',
      optimizer_switch              => '\'block_nested_loop=off\'',
      disable_log_bin               => '',
      mysqlx                        => 'off',
      log_error_verbosity           => '1'
    }
  },
}
```
Apologies - I guess this is likely a problem limited to the specific scenario (an internal module) where I'm using the module! Hopefully I'll work it out.

C24-AK:
> > > @C24-AK - Do you think you'd be able to add a `plugin` parameter to `$user_resource` of the `mysql::db` defined type, per my [earlier comment](https://github.com/puppetlabs/puppetlabs-mysql/pull/1612#issuecomment-1904503968), as part of this PR, or would you prefer to keep this separate?
> > 
> > 
> > Done, I appreciate if you could test the changes and see if everything works as expected
> 
> Thank you! I've looked at your changes and they look good. I'm attempting to test, but am facing a problem related to setting the mysql root password:
> 
> ```
> Error: can't modify frozen String: "ALTER USER 'root'@'localhost' IDENTIFIED WITH"
> Error: /Stage[main]/Mysql::Server::Root_password/Mysql_user[root@localhost]/password_hash: change from [old password hash redacted] to [new password hash redacted] failed: can't modify frozen String: "ALTER USER 'root'@'localhost' IDENTIFIED WITH"
> ```
> 
> I'm not sure if this problem relates to the changes in this PR, but I tried `15.0.0` from Puppet Forge and didn't see the problem. The Puppet resource relevant to the error is:
> 
> ```
> class {'::mysql::server':
>   package_name            => 'mysql-server',
>   service_name            => 'mysql',
>   remove_default_accounts => true,
>   restart                 => true,
>   service_enabled         => true,
>   config_file             => '/etc/mysql/my.cnf',
>   create_root_user        => true,
>   create_root_my_cnf      => true,
>   root_password           => $dlib_atom::mysql_root_password,
>   override_options        => {
>     mysqld => {
>       default-authentication-plugin => $mysql_user_auth_plugin,
>       sql_mode                      => 'ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION',
>       optimizer_switch              => '\'block_nested_loop=off\'',
>       disable_log_bin               => '',
>       mysqlx                        => 'off',
>       log_error_verbosity           => '1'
>     }
>   },
> }
> ```
> 
> Apologies - I guess this is likely a problem limited to the specific scenario (an internal module) where I'm using the module! Hopefully I'll work it out.

Hey, I appreciate testing this PR!
This error is unfortunately regarding how I concat the strings. Still don't know how ruby works :D
I didn't know I reference a frozen string which I cannot add a new string to.
I will check out the ruby docs  and see if I can fix this asap

Jimadine:
> Hey, I appreciate testing this PR! This error is unfortunately regarding how I concat the strings. Still don't know how ruby works :D I didn't know I reference a frozen string which I cannot add a new string to. I will check out the ruby docs and see if I can fix this asap

Thank you, your recent commit [d2e2b29](https://github.com/puppetlabs/puppetlabs-mysql/pull/1612/commits/d2e2b29c8a3a76e6ba2a36370844fa8ee9bb7c9a) fixed it! :smiley:

I guess it would mean yet another change to allow changing the authentication plugin for the root user? I think it would mean a new param in `server.pp` e.g. `root_plugin` and a change to `root_password.pp` - add new plugin attribute after line 35 e.g. `plugin => $mysql::server::root_plugin,`. What do you think?

C24-AK:
> > Hey, I appreciate testing this PR! This error is unfortunately regarding how I concat the strings. Still don't know how ruby works :D I didn't know I reference a frozen string which I cannot add a new string to. I will check out the ruby docs and see if I can fix this asap
> 
> Thank you, your recent commit [d2e2b29](https://github.com/puppetlabs/puppetlabs-mysql/pull/1612/commits/d2e2b29c8a3a76e6ba2a36370844fa8ee9bb7c9a) fixed it! 😃
> 
> I guess it would mean yet another change to allow changing the authentication plugin for the root user? I think it would mean a new param in `server.pp` e.g. `root_plugin` and a change to `root_password.pp` - add new plugin attribute after line 35 e.g. `plugin => $mysql::server::root_plugin,`. What do you think?

Good to hear that it works :)
Regarding plugin for root user, I noticed too that root password is always the same password. I can add a new parameter to that and make it adjustable.

Jimadine:
> @Jimadine I appreciate if you could test the root plugin :)

Thank you! I've tested it and it appears to be trying to change the plugin for root, but it's returning the following errors:
```
Error: Execution of '/usr/bin/mysql --database=mysql -e ALTER USER 'root'@'localhost' IDENTIFIED WITH 'caching_sha2_password' AS X'244124303035242816411A493F2969174D3D2D0825173421533D17662F4A6C687333456F484239596D4B43724D6A5375644964325673642E7074533052705436747943656B35'' returned 1: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
Error: /Stage[main]/Mysql::Server::Root_password/Mysql_user[root@localhost]/plugin: change from 'auth_socket' to 'caching_sha2_password' failed: Execution of '/usr/bin/mysql --database=mysql -e ALTER USER 'root'@'localhost' IDENTIFIED WITH 'caching_sha2_password' AS X'244124303035242816411A493F2969174D3D2D0825173421533D17662F4A6C687333456F484239596D4B43724D6A5375644964325673642E7074533052705436747943656B35'' returned 1: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
```

C24-AK:
> > @Jimadine I appreciate if you could test the root plugin :)
> 
> Thank you! I've tested it and it appears to be trying to change the plugin for root, but it's returning the following errors:
> 
> ```
> Error: Execution of '/usr/bin/mysql --database=mysql -e ALTER USER 'root'@'localhost' IDENTIFIED WITH 'caching_sha2_password' AS X'244124303035242816411A493F2969174D3D2D0825173421533D17662F4A6C687333456F484239596D4B43724D6A5375644964325673642E7074533052705436747943656B35'' returned 1: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
> Error: /Stage[main]/Mysql::Server::Root_password/Mysql_user[root@localhost]/plugin: change from 'auth_socket' to 'caching_sha2_password' failed: Execution of '/usr/bin/mysql --database=mysql -e ALTER USER 'root'@'localhost' IDENTIFIED WITH 'caching_sha2_password' AS X'244124303035242816411A493F2969174D3D2D0825173421533D17662F4A6C687333456F484239596D4B43724D6A5375644964325673642E7074533052705436747943656B35'' returned 1: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
> ```

I think I revert this commit since I cant actually test it right now, maybe we can implement this in another PR

Jimadine:
> I think I revert this commit since I cant actually test it right now, maybe we can implement this in another PR

Yeah, I agree, best to revert. FWIW, the same error was apparent when attempting to set root's plugin to `mysql_native_password`:
```
Error: Execution of '/usr/bin/mysql --database=mysql -e ALTER USER 'root'@'localhost' IDENTIFIED WITH 'mysql_native_password' AS '*B638EC5422004FCF44EE84FABA603D29A2259BC0'' returned 1: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
Error: /Stage[main]/Mysql::Server::Root_password/Mysql_user[root@localhost]/plugin: change from 'auth_socket' to 'mysql_native_password' failed: Execution of '/usr/bin/mysql --database=mysql -e ALTER USER 'root'@'localhost' IDENTIFIED WITH 'mysql_native_password' AS '*B638EC5422004FCF44EE84FABA603D29A2259BC0'' returned 1: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
```

C24-AK:
Good morning@bastelfreak, is there anything else to do for this PR?

C24-AK:
Any Reviewers available? @bastelfreak @alexjfisher

bastelfreak:
@C24-AK thanks for the patch! Is there a chance for you to add a test for this change?

nwton:
We can also replace this general regexp:
```
%r{0x[A-F0-9]+$}.match?(password)
```
with a more explicit one:
```
%r{0x24412430303524[A-F0-9]{40}$}.match?(password)
```

The magic sequence `24412430303524` is HEX for `$A$005$`, followed by 20 characters of SALT and 43 characters for the SHA DIGEST. It will remain in this format until some global changes are made to the MYSQL code.

C24-AK:
Hello,
I added the suggestions from @nwton.
But I am not really familiar with acceptance tests.
I would highly appreciate someone who could do it, so we can merge this PR asap :)

lib/puppet/provider/mysql_user/mysql.rb
```diff
@@ -224,7 +240,11 @@ def plugin=(string)
       end
     elsif newer_than('mysql' => '5.7.6', 'percona' => '5.7.6', 'mariadb' => '10.2.0')
       sql = "ALTER USER #{merged_name} IDENTIFIED WITH '#{string}'"
-      sql += " AS '#{@resource[:password_hash]}'" if string == 'mysql_native_password'
+      if string == 'mysql_native_password'
+        sql += " AS '#{@resource[:password_hash]}'"
+      elsif string == 'caching_sha2_password'
+        sql += " AS X'#{@resource[:password_hash][2..-1]}'"
+      end
```
bastelfreak, line 247: I think the ruby world prefers this, but I'm no expert:

```suggestion
      sql+= if string == 'mysql_native_password'
        " AS '#{@resource[:password_hash]}'"
      elsif string == 'caching_sha2_password'
        " AS X'#{@resource[:password_hash][2..-1]}'"
      end
```

bastelfreak:
@C24-AK can you please rebase against main to get rid of the merge commit?

lib/puppet/provider/mysql_user/mysql.rb
C24-AK: I had problems with the linting, I think it was regarding this but I am not sure anymore

C24-AK:
Push

bastelfreak:
@C24-AK any chance you can add a test for this?

C24-AK:
> @C24-AK any chance you can add a test for this?

As I mentioned above I am not familiar with acceptance tests and was looking forward for help here.
Otherwise I can read about it in the next days and try to figure out how to write these tests.

lib/puppet/functions/mysql/password.rb
```diff
@@ -22,7 +22,7 @@
   def password(password, sensitive = false)
     password = password.unwrap if password.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
 
-    result_string = if %r{\*[A-F0-9]{40}$}.match?(password)
+    result_string = if %r{\*[A-F0-9]{40}$}.match?(password) || %r{0x24412430303524[A-F0-9]{63}$}.match?(password)
```
binford2k, line 25: Would be great to get the explanation from https://github.com/puppetlabs/puppetlabs-mysql/pull/1612#issuecomment-2032406107 in here to make it more discoverable for future code archaeologists. Even better would be a link to the documented format, if someone happens to know it!  😁

```suggestion
    # This magic string is the hex encoded form of `$A$005${SALT}{SHA DIGEST}`, matching MySQL's expected format
    result_string = if %r{\*[A-F0-9]{40}$}.match?(password) || %r{0x24412430303524[A-F0-9]{63}$}.match?(password)
```

binford2k (changes requested):

lib/puppet/functions/mysql/password.rb
C24-AK: Hey thanks for the suggestion.
The only thing I can find is this blog https://dev.mysql.com/blog-archive/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/ and https://crypto.stackexchange.com/questions/77427/whats-the-algorithm-behind-mysqls-sha256-password-hashing-scheme

Format: DELIMITER[digest_type]DELIMITER[iterations]DELIMITER[salt][digest]

DELIMITIER: $
digest_type: A => SHA256
iterations: 005 => 5*ITERATION_MULTIPLIER
salt: Random string. Length SALT_LENGTH
digest: SHA2 digest. Length STORED_SHA256_DIGEST_LENGTH

CiderAndWhisky:
This has been going back and forth for 3/4 of a year - what is missing to get it merged? We use this branch on prod since December and need to always check to make sure not to overwrite it when we do module updates - would be nice to have it in the main branch. :-D

Thank you for the module!

lib/puppet/provider/mysql_user/mysql.rb
```diff
@@ -23,6 +23,13 @@ def self.instances
       # rubocop:enable Layout/LineLength
       @max_user_connections, @max_connections_per_hour, @max_queries_per_hour, @max_updates_per_hour, ssl_type, ssl_cipher,
       x509_issuer, x509_subject, @password, @plugin, @authentication_string = mysql_caller(query, 'regular').chomp.split(%r{\t})
+
+      if @plugin == 'caching_sha2_password'
+        # Escaping all single quotes to prevent errors when password generated it
+        @password = @password.gsub("'") { "\\'" }
+        @password = '0x' + @password.each_byte.map { |b| b.to_s(16) }.join
```
gmenuel, line 30: We've encountered idempotency problem with this line. The hash displayed by Puppet using `puppet resource mysql_user` doesn't match what is given by `SHOW CREATE USER`:

* the string given by `show create user` is all uppercase but in puppet it is all lowercase
* some `0` are not present when using puppet resource, I believe this is because the `to_s(16)` doesn't account for bytes lower than 16

I've been able to fix these two problems with the following code:

```suggestion
        @password = '0x' + @password.each_byte.map { |b| '%02X' % b.to_i }.join
```

C24-AK:
I tried to add tests to this, but unfortunately the instances are coming with mariadb.
Changing the packages does not work, since for example Debian 12 does not have mysql-server in its stable repository.
Since MariaDB Server does not provide support for caching_sha2_plugin, I cannot test this using acceptance tests.

I tested this in my own dev setup thoroughly.
Is there anything else I can provide?

CiderAndWhisky:
I will put a reminder for 2026 to see if this PR is still open, and another for 2035. My glass ball says 2026 it will be open and 2035 it will be rejected. Prove me wrong!

renzBeltran24:
Seems like some authtentication_string is not parsed correctly. For example we have set the password_hash = `0x244124303035241E47750A40457D30332E2D072D5F7F2801687A792F3948756137556371783878474C70385533456142454B464F2F384D306B58742F7A6B635859634C6C4131 `

However, it parsing it as: 
`0x24412430303524257C0C146D3A4D1411655C5C0447477C6954775A3B6730705350414577555478434E4A4C697768306552516A64644D56396F6E50397231584734593350786C44`

As you can see the auth string does not match and as a result, the module detects changes every time we run it.

Changing it to the following code works for us:
`@password = mysql_caller("SELECT CONCAT('0x',HEX('#{@password}'))", 'regular').chomp`

jochenbillen:
+1

daylicron:
From what I see this here seems to work but acceptance tests don't because they run on MariaDB. 

@bastelfreak @alexjfisher Do you have by chance any time to look at this? We really would like this feature, too.

sandwitch:
Hi,
Could you update on the status of this issue. I presume this is needed to use this module with caching_sha2_password?

C24-AK:
Thanks for the commit :)
I merged it into this one.

C24-AK:
I updated the PR:

Added plaintext password support for MySQL caching_sha2_password

We can now use plaintext passwords directly with the caching_sha2_password 
plugin instead of having to manually hash them. The code automatically 
detects plaintext passwords and hashes them using the proper SHA-256 crypt 
algorithm that MySQL expects.

*Note*: The implementation uses deterministic salt generation (based on password + 
username) so Puppet doesn't think the password changed on every run. This 
keeps everything idempotent while still being secure - different users get 
different salts even with the same password.

While working on this, I also found and fixed several SQL injection 
vulnerabilities where usernames weren't being properly escaped before going 
into SQL queries. This would not be an issue, since names are predetermined user inputs but better safe than sorry.

Also threw in some performance improvements.

The other plugins are unaffected by these changes.

Example usage:
  mysql::server::users:
    'test@localhost':
      plugin: caching_sha2_password
      password_hash: 'test' 

Tested on MySQL 8.4.7 in development environment.

danielkraaij:
Just wondering what the progress is on this PR?

jst-cyr:
@danielkraaij : Thanks for checking in on the PR. I've added this to our list to review and run through tests. I believe that right now this is waiting on Perforce Puppet engineering team to run through acceptance testing and to review the performance and security changes.

grant-veepshosting:
We operate a hosting fleet of ~150 MySQL nodes (Percona Server 8.0 and 8.4 LTS on Ubuntu jammy/noble, ~650 managed mysql_user resources) driven by this module, and we're about to adopt this PR as an in-tree patch on 16.3.0 to migrate our fleet off mysql_native_password — the provider's current rejection of non-* hashes is the only thing blocking us.

Two offers that might help move the pending review along:

Real-server acceptance results. Since the stated blocker has been that the acceptance CI only provisions MariaDB (which doesn't ship caching_sha2_password), we can run this branch through a staged pilot-then-promote rollout on actual Percona Server 8.0 and 8.4 hosts — create, plugin+hash change on existing users, idempotency across repeated runs, and post-restart full-authentication behaviour — and report the results back here. @jst-cyr would results from real MySQL-compatible servers (rather than the CI matrix) be useful input to the Perforce review, and is there a specific test checklist you'd want covered?

Idempotency check on the hex re-encode path. @renzBeltran24 reported (Apr 2025) that a stored hex hash was being re-parsed to a different value, causing perpetual change detection, and suggested using SELECT CONCAT('0x',HEX(...)) instead. Can @C24-AK confirm whether that fix (or the pure-Ruby equivalent @nwton suggested) is in the current branch? We'll specifically exercise this case in our pilot and report either way, since at hundreds of users any per-run churn would show up immediately in our reports.

Thanks @C24-AK for keeping this alive for 2.5 years — happy to help get it over the line.

C24-AK:
@grant-veepshosting
yes all changes should be on the current branch and works on production.
I can also confirm that multiple other teams are working with this branch in production.

Just for clarity. I don't think this will be merged anytime soon.
This is working and can be implemented and adjusted to your needs.
We migrated to openvox, since we think thats the future for opensource puppet.

jst-cyr:
@grant-veepshosting : I think the blocker on testing is more that the way the acceptance testing pipelines have been configured no longer work on PRs opened by community members. The secrets required to get access to Puppet Core and such are not released to forks. The engineering team has been working around this by creating their own PRs to run tests and get feedback, and then pass them back to the primary PR. 

It's not ideal and something we've been trying to fix.

grant-veepshosting:
Following up on our earlier comment — we finished the pilot. We ran the branch (as of `90321c6f`) against Percona Server 8.0 and 8.4 LTS on Ubuntu jammy/noble: user creation with `caching_sha2_password`, converting existing `mysql_native_password` accounts, idempotency across repeated runs, and re-reading existing hex hashes from the server all behaved correctly — including the idempotency case @renzBeltran24 flagged back in April 2025 (the hex re-encode path stayed stable across runs for us, no perpetual-change churn).

One real bug, though: the new function file `lib/puppet/functions/mysql/caching_sha2_password.rb` fails to load. It declares:

```ruby
Puppet::Functions.create_function(:caching_sha2_password) do
```

but because the file lives under the namespaced path `functions/mysql/`, Puppet's loader requires the namespaced function name and rejects it at call time with:

```
The code loaded from .../lib/puppet/functions/mysql/caching_sha2_password.rb produced mis-matched name, expected 'mysql::caching_sha2_password', got caching_sha2_password
```

The fix is a one-liner:

```ruby
Puppet::Functions.create_function(:'mysql::caching_sha2_password') do
```

We also found `mysql::db` doesn't benefit from this yet: it accepts the `plugin` parameter added earlier in this PR, but the `mysql_user` resource it builds still always hashes via `Deferred('mysql::password', [$password])` — i.e. native-only hashing — even when `plugin => 'caching_sha2_password'` is set. That leaves `mysql::db`-managed users with a native hash under a `caching_sha2_password` plugin, which the provider then mishandles. Small follow-up diff on top of this branch, in case it's useful to fold in:

```diff
diff --git a/manifests/db.pp b/manifests/db.pp
index 9247fb6c..8af90c3b 100644
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -104,9 +104,18 @@ define mysql::db (
   }
   ensure_resource('mysql_database', $dbname, $db_resource)
 
+  $db_password_hash = $plugin ? {
+    'caching_sha2_password' => Deferred('mysql::caching_sha2_password', [$password]),
+    default                 => Deferred('mysql::password', [$password]),
+  }
   $user_resource = {
     ensure        => $ensure,
-    password_hash => Deferred('mysql::password', [$password]),
+    password_hash => $db_password_hash,
     plugin        => $plugin,
     tls_options   => $tls_options,
   }
```

With the function-name fix applied, everything else worked for us as-is. We also confirmed the deterministic-salt hasher output authenticates successfully against Percona Server 8.4.10 for real client logins, not just stored on the account — so the hashing algorithm itself checks out.

Happy to provide more test detail, or open the `mysql::db` piece as a small PR against your branch directly if that's easier than a diff in a comment.

jst-cyr:
@C24-AK : We have some fixes for acceptance tests now available in the main branch which will be needed to run acceptance tests on this fix. Are you able to update the branch and resolve conflicts?
Also, if possible, would be great to do any squash/cleanup on the commit history. Thank you!

C24-AK:
@jst-cyr Thank you very much!
I implemented everything you mentioned.

Just in case someone is waiting, I cannot work on this PR anymore, so anyone could feel free to continue working on this.

jst-cyr:
I've done a cycle of running CI specs and resolving Rubocop and spec test issues. Spec tests against Puppet 8 open source are now green. Will run some platform acceptance tests next along with the Puppet Core 9 spec tests.

jst-cyr:
Running CI for this PR here: https://github.com/puppetlabs/puppetlabs-mysql/pull/1748

jst-cyr:
**Acceptance Test Results:** https://github.com/puppetlabs/puppetlabs-mysql/actions/runs/35783804138?pr=1748

Looks good! All green except for Debian 11 (both on 8 and 9). That's a failure that is also on main for the nightly, and recurring in other modules too. Not related to this PR.

Now that tests are all passing (and we have multiple confirmed in-the-wild production support for the solution) I think we just need some engineering approval on this to get it merged.

PR #1629, opened 2024-05-02
update PDK template and workflows

h0tw1r3:
- [x] https://github.com/puppetlabs/cat-github-actions/pull/88


.github/workflows/labeller.yml
```diff
@@ -2,26 +2,12 @@ name: Labeller
...
-
-      - uses: puppetlabs/community-labeller@v1.0.1
-        name: Label issues or pull requests
-        with:
-          label_name: community
-          label_color: '5319e7'
-          org_membership: puppetlabs
-          fail_if_member: 'true'
-          token: ${{ secrets.IAC_COMMUNITY_LABELER }}
+    if: contains(fromJson('["puppetlabs", "puppet-toy-chest"]'), github.repository_owner)
```
Ramesh7, line 11: do we really need `puppet-toy-chest` ?

.rubocop.yml
```diff
@@ -1,13 +1,11 @@
...
-inherit_from: .rubocop_todo.yml
-
 require:
 - rubocop-performance
 - rubocop-rspec
 AllCops:
   NewCops: enable
   DisplayCopNames: true
-  TargetRubyVersion: '2.7'
+  TargetRubyVersion: '2.6'
```
Ramesh7, line 8: This should be `2.7`

Ramesh7 (dismissed):
Reviewed the changes and looks good to me except minor cosmetic comments.

@bastelfreak @smortex can you please also have quick look?

.github/workflows/labeller.yml
h0tw1r3: yes, this is directly from https://github.com/puppetlabs/cat-github-actions/blob/main/example/module/labeller.yml

.rubocop.yml
h0tw1r3: No, this is managed by the pdk template, which currently specifies `2.6`.

Ramesh7 (dismissed):

.rubocop.yml
h0tw1r3: I would suggest opening a PR to the pdk-template project if you feel strongly about it.

types/package_ensure.pp
```diff
@@ -0,0 +1,2 @@
+# @summary Abstract for package_ensure values throughout the module
+type Mysql::Package_ensure = Variant[Enum['present','absent'], Pattern[/(\d+)[\.](\d+)[\.](\d+)/]]
```
smortex, line 2: Is it worth it to re-use the `Stdlib::Ensure::Package` data type rather than introducing a new one?

puppetlabs/puppetlabs-stdlib/5633b0645699a82fbf5d02ac1cec0ca72c0e1c2e/types/ensure/package.pp#L2

spec/classes/mysql_params_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'mysql::params' do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts }
+
+      it { is_expected.to compile }
```
smortex, line 10: I would expect a private class to not compile :smile: 

It happen that the class is marked as private in the API but does not call `assert_private()`…  Do we want to make sure private classes are not directly call and add `assert_private()` to all classes with `@api private`?  (I would say yes, as far as I am concerned, but have not checked if this can cause trouble).

smortex (commented):
Looks good overall.  I added inline comments for 2 tiny things that we may want to think about adjusting.

types/package_ensure.pp
h0tw1r3: makes sense to me

spec/classes/mysql_params_spec.rb
h0tw1r3: > I would expect a private class to not compile 😄 (the commit message of this change is about tests and private classes)

Removed

> It happen that the class is marked as private in the API but does not call assert_private()… Do we want to make sure private classes are not directly call and add assert_private() to all classes with @api private? (I would say yes, as far as I am concerned, but have not checked if this can cause trouble).

Makes sense to me, but this change is intended to be maintenance only, with no potentially breaking changes.
I'll open another PR for that.

h0tw1r3:
@Ramesh7 addressed issues, can you review and merge please?

jst-cyr:
Hello @h0tw1r3 , it looks like there was a lot of good work done on this a while back but it was never pulled up into the repo. The version of PDK being targeted in here is now rather old, so I would like to suggest we close off this PR.

I think we're going to need a new PR on this repo to update it to latest PDK and prepare it for Puppet Core 9.

jst-cyr:
For now I'm going to close this off and we can address with a separate Puppet Core 9 focus and the latest PDK.

PR #1634, opened 2024-06-30
(CAT-1919) - Handle scenario when user input password in <anything-in-caps-with-alpha-numeric>*<alpha-numeric-40-chars-in-caps>

Ramesh7:
## Summary

When the password is passed in `<anything-in-caps-with-alpha-numeric>*<alpha-numeric-40-chars-in-caps>` format with Sensitive then the current implementation fails with below error : 
```
Error: Only mysql_native_password (*ABCD...XXX) hashes are supported.
Error: /Stage[main]/Main/Mysql_user[test1@127.0.0.1]/password_hash: change from [old password hash redacted] to [new password hash redacted] failed: Only mysql_native_password (*ABCD...XXX) hashes are supported.
```
This is because the `mysql::password` doesn't handle mentioned condition properly.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/functions/mysql/password.rb
+++ b/lib/puppet/functions/mysql/password.rb
@@ -22,7 +22,7 @@ Puppet::Functions.create_function(:'mysql::password') do
   def password(password, sensitive = false)
     password = password.unwrap if password.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
 
-    result_string = if %r{\*[A-F0-9]{40}$}.match?(password)
+    result_string = if %r{^\*[A-F0-9]{40}$}.match?(password)
                       password
                     elsif password.empty?
                       ''
--- a/spec/acceptance/00_mysql_server_spec.rb
+++ b/spec/acceptance/00_mysql_server_spec.rb
@@ -63,9 +63,9 @@ describe 'mysql class' do
       end
 
       it 'can be set' do
-        apply_manifest(pp, catch_failures: true) do |r|
-          expect(r.stderr).to be_empty
-        end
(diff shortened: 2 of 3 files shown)


lib/puppet/functions/mysql/password.rb
```diff
@@ -22,7 +22,7 @@
   def password(password, sensitive = false)
     password = password.unwrap if password.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
 
-    result_string = if %r{\*[A-F0-9]{40}$}.match?(password)
+    result_string = if %r{^\*[A-F0-9]{40}$}.match?(password)
```
bastelfreak, line 25: I just have a few minutes right now, but I'm wondering why we have this regex at all? Does it check if it's a hash and otherwise runs into the last else block and runs SHA1?

alexjfisher: I think the intention is if it's already a Hash, then don't hash it, but since the regex isn't anchored it _can_ pick up other passwords.  The other password looks incredibly unlikely in the real world though.  But I guess @Ramesh7 bumped into this in their environment??

Ramesh7: @bastelfreak yeah, if its hash then it will return that otherwise it will hash it with SHA1.
@alexjfisher yeah, recently we encountered this and found there is flow in regex which needs to fix, so took it here.

bastelfreak (approved):

malikparvez (approved):

PR #1637, opened 2024-07-05
(MAINT) Removing EOL OS from metadata

Ramesh7:
## Summary

Removing EOL OS from metadata.

## Additional Context

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -23,13 +23,6 @@
       ]
     },
     {
-      "operatingsystem": "CentOS",
-      "operatingsystemrelease": [
-        "7",
-        "8"
-      ]
-    },
-    {
       "operatingsystem": "OracleLinux",
       "operatingsystemrelease": [
         "7"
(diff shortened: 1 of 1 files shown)


bastelfreak (changes requested):
please also remove the Code related to those OS versions

Ramesh7:
> please also remove the Code related to those OS versions

First can you please confirm the list is correct or there are more OS which is already EOL?

alexjfisher:
I know Voxpupuli's policy on this, but what is Puppet's? They still actively support EL7 in PE, don't they?

Ramesh7:
> I know Voxpupuli's policy on this, but what is Puppet's? They still actively support EL7 in PE, don't they?

EL7 you mean CentOS 7? If yes then yes. Have observed that PE team is working on to the same to remove it.
Also as the CentOS 7 EOL was on 30th June, so releasing new module with major version so I think its needed step as well.
Open for thoughts.

PR #1626, opened 2024-03-29
Add support for Mariadb 11.x

OxCom:
From MariaDB 10.4 we have deprecation notice that ``mysql*`` names will be removed. See: [JIRA ticket](https://jira.mariadb.org/browse/MDEV-29582)

Resolve #1580

Manually verified un Ubuntu 22.04 and MariaDB 11.x.

Deprecation notice examples:
```
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead.
mysqld: Deprecated program name. It will be removed in a future release, use '/usr/sbin/mariadbd' instead
mysqladmin: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb-admin' instead
```

PR #1644, opened 2024-08-28
Add support for Mariadb 11.x

### LINKED ISSUE #1580: Providers broken with MariaDB 11
## Describe the Bug

MariaDB 11 is out, and has deprecated the `mysql` (and `mysqldump` and `mysqladmin`) commands in favor of `mariadb` commands. The `mysql` commands now issue a deprecation message when used, which breaks the provider. e.g.:

```
Error: Could not prefetch mysql_database provider 'mysql': Execution of '/usr/bin/mysql -NBe show variables like '%_database' /usr/bin/mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead' returned 1: /usr/bin/mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
ERROR 1102 (42000): Incorrect database name '/usr/bin/mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/ma...'
```

And if we just run `mysql` directly on the command line:

```
# mysql
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 3927
Server version: 11.0.2-MariaDB Arch Linux

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]>
```

## Expected Behavior

The provider will work, without throwing errors because of the deprecation message. Ideally, the provider will detect the presence of `mariadb` (etc.) commands and prefer those over `mysql` commands (at least if the engine that's running is MariaDB). Or, failing that, the provider won't error out on the deprecation message.

## Steps to Reproduce

1. Install MariaDB 11.0.2 client.
2. Use any of the features in this module which call out to the `mysql` binary.

## Environment

 - Version: 15.0.0 (latest)
 - Platform: Arch Linux, Puppet 7.24.0

Marc-DRI:
## Summary
Add support for Mariadb 11.x and use its mariadb and mariadbd binaries.

## Additional Context
From Mariadb 11.0, mysql* names are deprecated in favor of mariadb* (cf. https://jira.mariadb.org/browse/MDEV-29582).
This pull request is:
- including commits from @OxCom, determining which binaries to use, depending on the mysqld_version facter
- fixing facters for mysql_version and mysqld_version whithin the MariaDB>11 context
- updating/adding tests for this new context

The initial PR #1626 causes [an infinite loop](https://github.com/puppetlabs/puppetlabs-mysql/pull/1626#issuecomment-2236002878) in our case of fresh deployment, because facter is not properly initialized.

## Related Issues
This pull request manages to fix #1580 .

### DIFF
--- a/.github/workflows/mend.yml
+++ b/.github/workflows/mend.yml
@@ -1,9 +1,10 @@
 name: "mend"
 
 on:
-  pull_request:
-    branches:
-      - "main"
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
--- a/lib/facter/mysql_version.rb
+++ b/lib/facter/mysql_version.rb
@@ -7,3 +7,11 @@ Facter.add('mysql_version') do
     mysql_ver.match(%r{\d+\.\d+\.\d+})[0] if mysql_ver
   end
 end
+
+Facter.add('mysql_version') do
+  confine { Facter::Core::Execution.which('mariadb') }
+  setcode do
+    mysql_ver = Facter::Core::Execution.execute('mariadb --version')
+    mysql_ver.match(%r{\d+\.\d+\.\d+})[0] if mysql_ver
+  end
+end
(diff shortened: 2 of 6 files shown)


Marc-DRI:
Will reopen a MR without commits behind.

PR #1645, opened 2024-08-28
Add support for Mariadb 11.x

### LINKED ISSUE #1580: Providers broken with MariaDB 11

Marc-DRI:
## Summary
Add support for Mariadb 11.x and use its mariadb and mariadbd binaries.

## Additional Context
From Mariadb 11.0, mysql* names are deprecated in favor of mariadb* (cf. https://jira.mariadb.org/browse/MDEV-29582).
This pull request is:
- including commits from @OxCom, determining which binaries to use, depending on the mysqld_version facter
- fixing facters for mysql_version and mysqld_version whithin the MariaDB>11 context
- updating/adding tests for this new context

The initial PR #1626 causes [an infinite loop](https://github.com/puppetlabs/puppetlabs-mysql/pull/1626#issuecomment-2236002878) in our case of fresh deployment, because facter is not properly initialized.

## Related Issues
This pull request manages to fix #1580 .

### DIFF
--- a/lib/facter/mysql_version.rb
+++ b/lib/facter/mysql_version.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
 Facter.add('mysql_version') do
-  confine { Facter::Core::Execution.which('mysql') }
   setcode do
-    mysql_ver = Facter::Core::Execution.execute('mysql --version')
+    mysql_ver = if Facter::Core::Execution.which('mysql')
+                  Facter::Core::Execution.execute('mysql --version')
+                elsif Facter::Core::Execution.which('mariadb')
+                  Facter::Core::Execution.execute('mariadb --version')
+                end
     mysql_ver.match(%r{\d+\.\d+\.\d+})[0] if mysql_ver
   end
 end
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -38,10 +38,28 @@ class Puppet::Provider::Mysql < Puppet::Provider
   ].join(':')
 
   # rubocop:disable Style/HashSyntax
-  commands :mysql_raw  => 'mysql'
-  commands :mysqld     => 'mysqld'
-  commands :mysqladmin => 'mysqladmin'
+  commands :mysql_client     => 'mysql'
+  commands :mariadb_client   => 'mariadb'
+  commands :mysqld_service   => 'mysqld'
+  commands :mariadbd_service => 'mariadbd'
+  commands :mysql_admin      => 'mysqladmin'
+  commands :mariadb_admin    => 'mariadb-admin'
   # rubocop:enable Style/HashSyntax
+
+  def self.mysql_raw(*args)
+    mysqld_version_string.scan(%r{mariadb}i) { return mariadb_client(*args) }
+    mysql_client(*args)
+  end
+
+  def self.mysqld(*args)
+    mysqld_version_string.scan(%r{mariadb}i) { return mariadbd_service(*args) }
+    mysqld_service(*args)
+  end
+
+  def self.mysqladmin(*args)
+    mysqld_version_string.scan(%r{mariadb}i) { return mariadb_admin(*args) }
+    mysql_admin(*args)
+  end
 
   # Optional defaults file
   def self.defaults_file
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -47,17 +47,23 @@ class Puppet::Provider::Mysql < Puppet::Provider
   # rubocop:enable Style/HashSyntax
 
   def self.mysql_raw(*args)
-    mysqld_version_string.scan(%r{mariadb}i) { return mariadb_client(*args) }
+    if newer_than('mariadb' => '11.0.0') && mysqld_version_string.scan(%r{mariadb}i)
+      return mariadb_client(*args)
+    end
     mysql_client(*args)
   end
 
   def self.mysqld(*args)
-    mysqld_version_string.scan(%r{mariadb}i) { return mariadbd_service(*args) }
+    if newer_than('mariadb' => '11.0.0') && mysqld_version_string.scan(%r{mariadb}i)
+      return mariadb_client(*args)
+    end
     mysqld_service(*args)
   end
 
   def self.mysqladmin(*args)
-    mysqld_version_string.scan(%r{mariadb}i) { return mariadb_admin(*args) }
+    if newer_than('mariadb' => '11.0.0') && mysqld_version_string.scan(%r{mariadb}i)
+      return mariadb_client(*args)
+    end
     mysql_admin(*args)
   end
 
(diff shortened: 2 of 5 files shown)


lib/facter/mysql_version.rb
```diff
@@ -7,3 +7,11 @@
     mysql_ver.match(%r{\d+\.\d+\.\d+})[0] if mysql_ver
   end
 end
+
+Facter.add('mysql_version') do
```
bastelfreak, line 11: now we've two blocks for the same fact. Doesn't it make sense to unify it? What if mysql and mariadb is installed on the same server?

smortex: This documentation feels weird.  Inspecting various facts in the puppetlabs organization repositories, I don't see usage of confine with different prerequisites for the same facts, but a single fact with confinement that checks multiple conditions, e.g. https://github.com/puppetlabs/puppetlabs-java/blob/6319799bb310f824b18d9077905c46409535c977/lib/facter/java_default_home.rb#L16

It feels more natural for me to have a single `Facter.add('mysql_version')`, where the setcode logic attempts to get the expected value and return nil if none was found.  A confinement does not even provide a short-circuit in this case, so we can probably go without it, i.e.

```ruby
Facter.add('mysql_version') do
  setcode do
    if Facter::Core::Execution.which('mysql')
      mysql_ver = Facter::Core::Execution.execute('mysql --version')
      mysql_ver.match(%r{\d+\.\d+\.\d+})[0] if mysql_ver
    elsif Facter::Core::Execution.which('mariadb')
      mysql_ver = Facter::Core::Execution.execute('mariadb --version')
      mysql_ver.match(%r{\d+\.\d+\.\d+})[0] if mysql_ver
    end
  end
end
```

So merging may make sense IMHO.


But beyond this consideration (and speaking as someone who does not use mysql nor mariadb), if mariadb and mysql are diverging one from the other and could be installed side by side on a single system, maybe it make sense to "fork" the puppetlabs-mysql module (puppet(labs)?-mariadb) just like mariadb forked mysql so that each module can track what the software they configure implement and provide a consistent interface and not some weak abstraction that may or may not work depending on the package you manage with the module?

Marc-DRI: Facters reworked and correctly working.
Thank you @smortex , code is much nicer !

neomilium (approved):

neomilium:
> But beyond this consideration (and speaking as someone who does not use mysql nor mariadb), if mariadb and mysql are diverging one from the other and could be installed side by side on a single system, maybe it make sense to "fork" the puppetlabs-mysql module (puppet(labs)?-mariadb) just like mariadb forked mysql so that each module can track what the software they configure implement and provide a consistent interface and not some weak abstraction that may or may not work depending on the package you manage with the module?

@smortex I agree. It could be the right moment to split modules : `mariadb` and `mysql`.

@bastelfreak : What is your through about this split?

neomilium:
Thinking it twice, there is no much interest to fork and no sufficient divergence to need a fork, so IMHO this PR could be accepted as is.

smortex (approved):
Looks good to me.  I am not a member of this project so cannot trigger a CI run, @bastelfreak @alexjfisher can you please trigger it?  Thanks!

bastelfreak:
Hey,
I'm currently waiting with most work on the Perforce owned modules until https://www.puppet.com/blog/open-source-puppet-updates-2025 is resolved.

lib/puppet/provider/mysql.rb
```diff
@@ -38,11 +38,35 @@ class Puppet::Provider::Mysql < Puppet::Provider
...
+  def self.mysql_raw(*args)
+    if newer_than('mariadb' => '11.0.0') && mysqld_version_string.scan(%r{mariadb}i)
+      return mariadb_client(*args)
+    end
+    mysql_client(*args)
+  end
+
+  def self.mysqld(*args)
+    if newer_than('mariadb' => '11.0.0') && mysqld_version_string.scan(%r{mariadb}i)
+      return mariadb_client(*args)
```
ekohl, line 58: Is this really correct? With `mysqld` I'd have expected some server command

lib/puppet/provider/mysql.rb
```diff
@@ -38,11 +38,35 @@ class Puppet::Provider::Mysql < Puppet::Provider
   ].join(':')
 
   # rubocop:disable Style/HashSyntax
-  commands :mysql_raw  => 'mysql'
-  commands :mysqld     => 'mysqld'
-  commands :mysqladmin => 'mysqladmin'
+  commands :mysql_client     => 'mysql'
+  commands :mariadb_client   => 'mariadb'
```
ekohl, line 42: You should probably use `optional_commands` to be explicit that they're not expected to be present (or the `has_command` definition together with `is_optional`).

lib/puppet/provider/mysql.rb
```diff
@@ -38,11 +38,35 @@ class Puppet::Provider::Mysql < Puppet::Provider
...
+  commands :mariadbd_service => 'mariadbd'
+  commands :mysql_admin      => 'mysqladmin'
+  commands :mariadb_admin    => 'mariadb-admin'
   # rubocop:enable Style/HashSyntax
 
+  def self.mysql_raw(*args)
+    if newer_than('mariadb' => '11.0.0') && mysqld_version_string.scan(%r{mariadb}i)
+      return mariadb_client(*args)
+    end
+    mysql_client(*args)
```
ekohl, line 53: Also, I think we can assume Ruby 2.7 and use argument forwarding.

Once you've defined them as optional, I wonder if this works:

```suggestion
  def self.mysql_raw(...)
    command(:mariadb_client) ? mariadb_client(...) : mysql_client(...)
```

lib/puppet/provider/mysql.rb
```diff
@@ -38,11 +38,35 @@ class Puppet::Provider::Mysql < Puppet::Provider
   ].join(':')
```
ekohl, line 38: I'm looking at this horrible hack above and wonder if it should use the proper API. Something like this probably works:
```ruby
has_command(:mysql_client, 'mysql') do
  environment PATH: path, LD_LIBRARY_PATH: ld_library_path
end
```

You can also use `is_optional` according to https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/lib/puppet/provider.rb#L177-L214.

imaqsood (approved):

danir-de:
Can this be merged soon?

PR #1659, opened 2024-12-11
fix: manage directory containing log

vchepkov:
Currently, /var/log/mysql is created on all mysql servers regardless if they are used or not
This change instead manages directory where configured log file is configured


vchepkov:
@shubhamshinde360 , I think this was the intent of your change

PR #1694, opened 2025-10-17
Use params::provider instead of "mysql" to import sql

teluq-pbrideau:
## Summary

Use `$mysql::params::provider` instead of `mysql` to import a database in `mysql::db`.

As this parameter is already configured for some OS, I decided to define this variable for every OS, and use it.

## Additional Context
See #1693

## Related Issues (if any)
#1693

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/db.pp
+++ b/manifests/db.pp
@@ -124,7 +124,7 @@ define mysql::db (
 
     if $sql {
       exec { "${dbname}-import":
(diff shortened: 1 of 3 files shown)


bronachfalls:
Hi @teluq-pbrideau , could you please fix the syntax errors highlighted in the Spec tests? Thanks!

bronachfalls (approved):

Issue #1673, opened 2025-06-14
incorrectly fails support checks for Enterprise Linux 10


ikonia:
## Describe the Bug
using 16.2.0 puppetlabs-mysql module on Enterprise Linux 10 (multiple variants) the basic functionality incorrectly fails version compatibility checks when prepping to install database managed packages

puppetlabs-mysql only supports RedHat 7.0 and beyond.

## Expected Behavior
module executes basic functionality during a puppet run and completes successfully (test was install rpm packages to support mysql/mariadb)

## Steps to Reproduce
1. Install Enterprise Linux 10 operating system (RedHat / Rocky / Alma) 
2. Install OpenVox puppet agent 8 (any version aligned to the 8 release) 
3. include puppetlabs-mysql 16.2.0 on your puppet master modules
4. include mysql::server manifest with a basic set of parameters (inline with module example)
5. execute puppet run

## Environment
 - Version 16.2.0
 - Platform Rocky Linux 10.0
 - OpenVox openvox-agent-8.19.2-1
 - 
## Additional Context
Puppet run fails with 

`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Unsupported platform: puppetlabs-mysql only supports RedHat 7.0 and beyond. (file: /etc/puppetlabs/code/environments/production/modules/mysql/manifests/params.pp, line: 474, column: 5) `

the version check looks like 

` ## Additional graceful failures
  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] < '7' and $facts['os']['name'] != 'Amazon' {
    fail("Unsupported platform: puppetlabs-${module_name} only supports RedHat 7.0 and beyond.")
`

however the facter output shows this should pass 

`[root@king consul]# facter os.family
RedHat`

`[root@king consul]# facter os.release.major
10`

timmooney:
It's because they're doing string comparison rather than using versioncmp(), and when comparing strings '10' is less than '7' because of the first character of each.

Other parts of the module are already using versioncmp() , this just needs to be updated to use it too.

davemcdonnell:
RHEL10 approaching 1 year GA and this still isnt fixed?

PR #1712, opened 2026-06-30
(MODULES-11802) Add support for RHEL 10

skyamgarp:
## Summary
Adds Red Hat Enterprise Linux 10 to the list of supported operating systems.

The only change required is declaring `RedHat` `10` in `metadata.json`'s `operatingsystem_support`. The version-branching logic in `manifests/params.pp` already uses `versioncmp` for every RedHat-family comparison, so RHEL 10 falls into the existing `>= 8` branches and resolves to the **mariadb** provider. This complements the earlier `versioncmp` fix for CentOS Stream 10 (commit 68fcb4c, #1673).

## Changes
- `metadata.json`: add `"10"` to RedHat `operatingsystemrelease`.

## Verification
- `bundle exec rake spec_prep` + `mysql_server`/`mysql_client`/`mysql_backup_xtrabackup` specs: **950 examples, 0 failures**.
- Synthetic RHEL 10 catalog compiles cleanly (`compile.with_all_deps`) and correctly resolves to the mariadb provider:
  - `Package[mysql-server]` → `name => mariadb-server`
  - `Service[mysqld]` → `name => mariadb`
  - config `/etc/my.cnf.d`, log `/var/log/mariadb/mariadb.log`

## Notes
- `on_supported_os` does not yet auto-generate a RHEL 10 context because FacterDB has no RHEL 10 fact set in the current gem version (same limitation as the CentOS Stream 10 fix); the suite continues to exercise 8/9.
- Acceptance targets (`provision.yaml`) were intentionally left unchanged, matching the scope of the CentOS Stream 10 change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/puppet/type/mysql_database.rb
+++ b/lib/puppet/type/mysql_database.rb
@@ -17,9 +17,11 @@ Puppet::Type.newtype(:mysql_database) do
   # `utf8` is a deprecated alias for `utf8mb3` in both MySQL and MariaDB. Newer
   # servers (e.g. MariaDB 11 on RHEL 10) report the canonical `utf8mb3` name back,
   # while users typically request `utf8`. Treat the two spellings as equivalent so
-  # the resource stays idempotent regardless of which the server reports.
+  # the resource stays idempotent regardless of which the server reports. Charset
+  # and collation names are case-insensitive and the server reports them
+  # lowercased, so downcase first to also match user input such as `UTF8`.
   def self.normalise_utf8(value)
-    value.to_s.sub(%r{^utf8mb3}, 'utf8')
+    value.to_s.downcase.sub(%r{^utf8mb3}, 'utf8')
   end
 
   newproperty(:charset) do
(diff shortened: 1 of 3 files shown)


skyamgarp:
The PR check failures are existing and can be addressed separately.

lib/puppet/type/mysql_database.rb
```diff
@@ -14,15 +14,31 @@
     desc 'The name of the MySQL database to manage.'
   end
 
+  # `utf8` is a deprecated alias for `utf8mb3` in both MySQL and MariaDB. Newer
+  # servers (e.g. MariaDB 11 on RHEL 10) report the canonical `utf8mb3` name back,
+  # while users typically request `utf8`. Treat the two spellings as equivalent so
+  # the resource stays idempotent regardless of which the server reports.
+  def self.normalise_utf8(value)
+    value.to_s.sub(%r{^utf8mb3}, 'utf8')
```
mehul-jain1: should we also add downcase here as well if user pass value in uppercase 

```suggestion
    value.to_s.downcase.sub(%r{^utf8mb3}, 'utf8')
```

skyamgarp: Hi Mehul, added the suggested fix, can you please take a look at it?

mehul-jain1 (approved):

PR #1716, opened 2026-07-09
Respect ensure when managing xtrabackup backup package

milad-zanganeh:
## Summary
Fixes `mysql::backup::xtrabackup` so that the backup method package and backup directory are removed when the class is declared with `ensure => absent`.

## Additional Context
- [x] Root cause and the steps to reproduce.
  - Root cause: `stdlib::ensure_packages($backupmethod_package)` was called without an options hash, defaulting the package to `ensure => present` regardless of the class `$ensure` parameter.
  - The $backupdir file resource was hardcoded to ensure => 'directory', which ignores the class's $ensure parameter.
  - Reproduce: Declare `mysql::server::backup` (with `backupmethod => 'xtrabackup'`) and `ensure => absent`, run `puppet apply`, and observe that the backup method package remains installed and still manages the backup directory.
- [x] Thought process behind the implementation.
  - The class already exposes an `$ensure` parameter (`Enum['present', 'absent']`) used for the other backup resources. Passing it through to `stdlib::ensure_packages` and file makes the package and directory lifecycle consistent with the rest of the backup resources without introducing a new parameter.

### DIFF
--- a/manifests/backup/xtrabackup.pp
+++ b/manifests/backup/xtrabackup.pp
@@ -35,7 +35,7 @@ class mysql::backup::xtrabackup (
   String[1]                                     $backupmethod_package     = $mysql::params::xtrabackup_package_name,
   Array[String]                                 $excludedatabases = [],
 ) inherits mysql::params {
-  stdlib::ensure_packages($backupmethod_package)
+  stdlib::ensure_packages($backupmethod_package, { 'ensure' => $ensure })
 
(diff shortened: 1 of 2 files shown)


jst-cyr:
Thanks for the submission @milad-zanganeh ! First glance by me looks like this is a good addition, I'm going to add it to the list for engineering to review.

bronachfalls (approved):

PR #1717, opened 2026-07-14
Squelch assorted test complaints

greatflyingsteve:
When running module tests locally, random issues with the test framework, some of which would have invalidated the tests they were part of, caused almost as many lines of error spam during a test run as a clean test run itself would have produced.  Fix these aggravating leftover issues and assorted deprecations so the tests are not as annoying to use.

## Summary
- Remove deprecated `should` syntax in favor of `expect` to squelch warnings like `Deprecation Warnings:  Using 'should' from rspec-expectations' old ':should' syntax without explicitly enabling the syntax is deprecated.`
- Fix a broken Facter mock so it no longer produces 45 lines of warnings like `WARNING: An expectation of ':root_home' was set on 'nil'. To allow expectations on 'nil'...` by attempting to mock methods that do not exist
- Top-level fact names are symbols; stop merging string names into the facts hash and merge symbols instead.  The string names do not merge with the symbols they replace, but they do take precedence over symbol names during retrieval; so all of the data attached to the existing symbol-named fact is effectively lost, and you get fun errors like `Could not retrieve fact networking.ip`
- Convert legacy facts-as-top-scope-variable references in acceptance tests, since these all go away as of Puppet 8.  These did not produce error output that I could see, since I can't run the acceptance tests in my environment, but came up while looking for the `root_home` culprit above.

## Checklist
- [x] 🟢 Spec tests.

### DIFF
--- a/spec/acceptance/types/mysql_grant_spec.rb
+++ b/spec/acceptance/types/mysql_grant_spec.rb
@@ -282,7 +282,7 @@ describe 'mysql_grant' do
 
         exec { 'mysql-create-table':
           command     => '/usr/bin/#{mysql_cmd} -NBe "CREATE TABLE foo.bar (name VARCHAR(20))"',
-          environment => "HOME=${::root_home}",
+          environment => "HOME=${facts['root_home']}",
           unless      => '/usr/bin/#{mysql_cmd} -NBe "SELECT 1 FROM foo.bar LIMIT 1;"',
           require     => Mysql_database['foo'],
         }
@@ -334,12 +334,12 @@ describe 'mysql_grant' do
           user       => "web@${dbSubnet}",
           require    => Mysql_user["web@${dbSubnet}"],
         }
-        mysql_user { "web@${::networking['ip']}":
+        mysql_user { "web@${facts['networking']['ip']}":
           ensure => present,
         }
-        mysql_grant { "web@${::networking['ip']}/*.*":
-          user       => "web@${::networking['ip']}",
-          require    => Mysql_user["web@${::networking['ip']}"],
+        mysql_grant { "web@${facts['networking']['ip']}/*.*":
+          user       => "web@${facts['networking']['ip']}",
+          require    => Mysql_user["web@${facts['networking']['ip']}"],
         }
         mysql_user { 'web@localhost':
           ensure => present,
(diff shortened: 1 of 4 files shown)


jst-cyr:
@greatflyingsteve : Do you happen to know if the changes you've made here would alter any lower bounds of requirements in the module? It looks great to get rid of some warnings noise and modernize a little, I'm just trying to get an idea of the scope of impact on existing users (i.e. would something like this require a major version release)

greatflyingsteve:
> @greatflyingsteve : Do you happen to know if the changes you've made here would alter any lower bounds of requirements in the module? It looks great to get rid of some warnings noise and modernize a little, I'm just trying to get an idea of the scope of impact on existing users (i.e. would something like this require a major version release)

@jst-cyr  I'm confident it wouldn't alter any lower bounds at all.  The changes here are all either shifting away from deprecated methods in RSpec itself, or shifting _toward_ paradigms or features in either Puppet or RSpec-Puppet that have been best practice since Puppet 6 or earlier (e.g. using Symbol fact names, using the `facts` hash vs. top-scope variables).

jst-cyr:
Thank you @greatflyingsteve ! This is good for the engineering team to know when they are reviewing.

jst-cyr:
Running CI tests for this PR here: https://github.com/puppetlabs/puppetlabs-mysql/pull/1729

jst-cyr:
Acceptance test results: https://github.com/puppetlabs/puppetlabs-mysql/actions/runs/31841390137/job/94898936307?pr=1729

Acceptance test results compared against the latest nightly run against `main`:

**All failures in PR acceptance testing match existing failures in the Nightly build against main.**

| Platform | PR | Nightly | Failure |
|---|---|---|---|
| SLES-12 | ❌ failure | ❌ failure | Fails during `Provision environment` — the sles-12 node provisions successfully, but the subsequent SSH connection to it fails (`Errno::ECONNREFUSED` / timeout on port 22) after 3 retries |

All other 13 platforms (Ubuntu-18.04/22.04, RedHat-7/8/9/10, Debian-11/12, Rocky-8, SLES-15, CentOS-8) passed.

The SLES-12 failure is a pre-existing infrastructure/provisioning issue on `main` (SSH unreachable after provisioning), not something introduced by this PR, no action needed here.

Still require engineering review, but at least we have acceptance test feedback now!

bronachfalls (approved):

PR #1695, opened 2025-11-12
Add purge and recurse options to the include directory creation when …

bschonec:
…MariaDB package.

## Summary
On MariaDB package, the purge/recurse options weren't set for the include_dir directory.

PR #1730, opened 2026-08-17
Add purge and recurse options to the include directory creation when MariaDB package.

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-mysql/pull/1695)


jst-cyr:
All green, except for SLES 12, which matches the results on nightly main. Feedback provided on original PR.

PR #1735, opened 2026-08-24
fix: detect mariadb binary availability on RedHat family

bronachfalls:
## Problem

When `mysql::db` is used with the `sql` parameter to import a SQL file, the module runs an `Exec` resource to pipe the file into the database client. Since `bb99a9b`, the client command is determined by `$mysql::params::provider`, which was hardcoded to `'mariadb'` for all RedHat-family systems on OS version 7+.

This causes a failure on RedHat-family systems where only the `mysql` binary is available (not `mariadb`):

```
Error: /Stage[main]/Main/Mysql::Db[spec2]/Exec[spec2-import]: sh: mariadb: command not found
```

Confirmed failing on:
- `litmusimage/rockylinux:8`
- `litmusimage/centos:stream8`

And in CI against RedHat 7, RedHat 8, CentOS 8, and Rocky 8.

## Root Cause

The MariaDB project renamed its client binary from `mysql` to `mariadb` starting in MariaDB 11. However, not all RedHat-family systems ship MariaDB 11+ — many still only provide the `mysql` binary. Hardcoding `provider = 'mariadb'` for the entire RedHat family assumes all systems have the newer binary, which is incorrect.

## Fix

Instead of hardcoding the provider by OS version, `params.pp` now uses Puppet's `find_file()` function to detect which binary is actually present on the system at catalog compile time:

```puppet
$provider = find_file('/usr/bin/mariadb', '/bin/mariadb') ? {
  undef   => 'mysql',
  default => 'mariadb',
}
```

This mirrors the approach already used in `spec/spec_helper_acceptance_local.rb`:

```ruby
def get_db_cmd
  run_shell('mariadb -V', expect_failures: true).stdout.empty? ? 'mysql' : 'mariadb'
end
```

Both follow the same principle: check what is actually available rather than assuming based on OS version.

## Unit Spec Tests

`find_file()` runs at catalog compile time against the real filesystem of the machine compiling the catalog. In unit tests, that machine is the Ubuntu CI runner, which has no `mariadb` binary. `find_file()` therefore returns `undef` and `$provider` falls back to `'mysql'`.

The unit spec for `mysql::db` previously hardcoded `'mariadb'` as the expected provider for all RedHat-family OS contexts, which no longer matches. The expectation has been updated to `'mysql'` to reflect what `find_file()` produces in the unit test environment. The `mariadb` path is covered by acceptance tests, which run against real RedHat containers where the binary is present after package installation.

## Impact

- Systems with MariaDB 11+ (which provide the `mariadb` binary) continue to work as before
- Systems with older MariaDB or MySQL (which only provide the `mysql` binary) now work correctly
- No change to behaviour on non-RedHat platforms

### DIFF
--- a/lib/facter/mysql_binary.rb
+++ b/lib/facter/mysql_binary.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+Facter.add('mysql_binary') do
+  setcode do
+    if Facter::Core::Execution.which('mariadb')
+      'mariadb'
+    elsif Facter::Core::Execution.which('mysql')
+      'mysql'
+    end
+  end
+end
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -44,9 +44,9 @@ class mysql::params {
         }
         /^(RedHat|Rocky|CentOS|Scientific|OracleLinux|AlmaLinux)$/: {
           if versioncmp($facts['os']['release']['major'], '7') >= 0 {
-            $provider = find_file('/usr/bin/mariadb', '/bin/mariadb') ? {
-              undef   => 'mysql',
-              default => 'mariadb',
+            $provider = $facts['mysql_binary'] ? {
+              undef   => 'mariadb',
+              default => $facts['mysql_binary'],
             }
             if versioncmp($facts['os']['release']['major'], '8') >= 0 {
               $xtrabackup_package_name = 'percona-xtrabackup-24'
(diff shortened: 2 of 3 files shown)


gavindidrichsen (changes requested):
The switch in this PR should be done with a fact and not a find_files.  The reason for this is that puppet catalogs are created on the primary or compiler not on the target node.  In other words, you might have a target node on which you want MYSQL.  If, however, the primary has a /usr/bin/mariadb when the target does not, then the catalog compilation will do a find_files  on the primary, find the /usr/bin/mariadb .  The target node will fail again because there is no /usr/bin/mariadb.

PR #1713, opened 2026-07-01
Fix `mysql_*` providers for MariaDB >= 11.4

cruelsmith:
## Summary
Fixes the broken providers for MariaDB >= 11.4 as already tried by #1645 but now working by using `optional_commands` and `confine` block so that provider suitability check works as exspected.

The `confine` uses `Puppet::Util.which()` as the `optional_commands` can not be used at the time for the provider suitability check.
On usage the command (wrapper around which) finds the first existing binary and we also caches it for later calls and provider instances.

Style disclaimer: The normal way would have been to copy `provider/mysql_user/mysql.rb` to `provider/mysql_user/mariadb.rb` and do the needed changes for mariadb like `commands mysql_raw: 'mariadb'`. This change keeps the idea of #1645 to not have multiple files for this.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Closes #1672
Closes #1685 as duplicate of #1672
Closes #1689

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified. (via puppet on RHEL9 with MariaDB 12.3)

PR #1739, opened 2026-09-04
Fix `mysql_*` providers for MariaDB >= 11.4

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-mysql/pull/1713)

### DIFF
--- a/lib/puppet/provider/mysql.rb
+++ b/lib/puppet/provider/mysql.rb
@@ -37,34 +37,20 @@ class Puppet::Provider::Mysql < Puppet::Provider
     '/usr/mysql/5.7/lib64',
   ].join(':')
 
-  # rubocop:disable Style/HashSyntax
-  commands :mysql_client     => 'mysql'
-  commands :mariadb_client   => 'mariadb'
-  commands :mysqld_service   => 'mysqld'
-  commands :mariadbd_service => 'mariadbd'
-  commands :mysql_admin      => 'mysqladmin'
-  commands :mariadb_admin    => 'mariadb-admin'
-  # rubocop:enable Style/HashSyntax
+  COMMANDS = {
+    mysql_raw: ['mysql', 'mariadb'],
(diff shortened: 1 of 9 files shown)


cruelsmith:
Aha i see the `fails to execute while applying` with `pp_three` are failing as they try to temp remove the mysql binary but now we can also use the mariadb binary 😓
As this fail the `pp_three` is left broken the other test after that fail also.

jst-cyr:
Debian 11 results are strange... Going to rerun that one and see if we can't get it to run the tests. I suspect it will see the same issues, though.

jst-cyr:
**Acceptance Test Results:** https://github.com/puppetlabs/puppetlabs-mysql/actions/runs/33911527207?pr=1739

You're read on the error messages seems to be consistent across the failing systems. 

[mysql_grant_spec.rb:638](vscode-webview://02t8u8ps7aq2bfc2baoniq0ro7skjcuv27ip5sk623idsiua6e9t/spec/acceptance/types/mysql_grant_spec.rb#L638) has a test that renames the mysql binary away, then expects Puppet to fail with "Could not find a suitable provider for mysql_grant." 

Before your change, that was true. Now that the provider also accepts mariadb as a valid command, it's still "suitable" with mysql gone — so Puppet actually runs the grant via the mariadb binary instead of failing, and the test's expect(...).to match(...) assertion fails.

Because that assertion is the second-to-last line in the test, the last line — which renames mysql back — never runs. mysql stays missing for the rest of the suite, so the next two tests fail too, both with a plain "mysql: not found."

All 6 platforms (Ubuntu-22.04, Debian-11, Debian-12, RedHat-9, RedHat-10, SLES-15) hit this exact same 3 failures, at the same spec lines. Confirmed it's just this one issue.

Feedback will be provided on the original PR and this one will be closed.

PR #1744, opened 2026-09-16
Fix `mysql_*` providers for MariaDB >= 11.4

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-mysql/pull/1713)


cruelsmith:
🤦 i should have not changed spec_helper_acceptance_local.rb with 934ba4ebdcdae2a80dc73bfa40e11b75cfd988bc. Without it would work...

jst-cyr:
**Acceptance Test Results:** https://github.com/puppetlabs/puppetlabs-mysql/actions/runs/35141835394?pr=1744

It looks like the PR has broken all of the acceptance tests. I have rerun a few times, in case it was a provisioning or infrastructure flakiness issue, but it seems to be fairly consistently broken.

Feedback will be copied back to the original PR.

